PR #6, opened 2015-06-06
Replaced strings with Android's native labels.

MacroYau:


MacroYau:
Just found that "resource is not public". Consider copying these strings from the SDK for translation.

PR #14, opened 2015-06-10
Update AndroidManifest.xml

sleepingAnt:
Make MainActivity "singleTask",otherwise there will be too much instance.


PaoloRotolo:
Looks good. Thank you!

PR #18, opened 2015-06-13
Removed second fragment instantiation in addSlide() to support instantiation with arguments

leon1101:
Example for fragment instantiation with arguments:

```java
public static class SlideFragment extends Fragment {

    private static final String ARG_LAYOUT_RES_ID = "layoutResId";

    public static SlideFragment newInstance(int layoutResId) {
        SlideFragment slideFragment = new SlideFragment();

        Bundle args = new Bundle();
        args.putInt(ARG_LAYOUT_RES_ID, layoutResId);
        slideFragment.setArguments(args);

        return slideFragment;
    }

    private int layoutResId;

    public SlideFragment() {}

    @Override
    public void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        if(getArguments() != null && getArguments().containsKey(ARG_LAYOUT_RES_ID))
            layoutResId = getArguments().getInt(ARG_LAYOUT_RES_ID);
    }

    @Nullable
    @Override
    public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
        return inflater.inflate(layoutResId, container, false);
    }

}
```

```java
addSlide(SlideFragment.newInstance(R.layout.fragment_slide_one);
```


girish3:
+1 
But rather than directly adding the passed fragment to the list, it should instantiate using the given context
```java
fragments.add(fragment);
// instead of above line it should be below line
fragments.add(Fragment.instantiate(context, fragment.getClass().getName(), fragment.getArguments()));
// it won't matter even if the Bundle object is null 
```

PaoloRotolo:
@girish3, great. Can you guys please update the pull request? :+1:

girish3:
@PaoloRotolo May I ask why instantiate method was used rather than adding the fragment directly to the list?

leon1101:
@girish3 I don't think that's necessary here since  instantiate() is called with the same context.

leon1101:
@girish3 I'd like to wait for @PaoloRotolo 's answer before I update my pull request.

PaoloRotolo:
@girish3 yes, I think context will be the same. Thanks in advance for the pull :+1:

PaoloRotolo:
Thank you!

PR #23, opened 2015-06-20
Changed project structure to import easier

porokoro:
The actual project layout has some redundant files and is missing things like global .gitignore files. In addition it is difficult to import the whole projekt at once into IntelliJ or Android Studio.

At the moment you need to import the library and the example as separate projects which is somewhat incovenient.

Therefore I changed the structure so that you can import it easily.

Also I updated the gradle wrapper to the newest version.

The second commit is for the missing method in AppIntro2. Because of the missing method the project could not compile.


PaoloRotolo:
Great work, thank you!

PR #25, opened 2015-06-21
Moved FadePageTransformer and PagerAdapter into separate files

porokoro:
The ``FadePageTransformer`` and ``PagerAdapter`` where exactly the same inside ``AppIntro`` and ``AppIntro2``. I just moved them into separate files so there is no need to define them twice.

**Warning**: Both classes are now _package local_ instead of _public_ so this could be an API breaker. But there should be no reason to derive from this classes because the are intended for internal use only in my opinion. 
To take the safe route consider bumping the version number to 2.0.0 for the next release if you use [semantic versioning](http://semver.org/).


PaoloRotolo:
Nice work again. I think I'll release a new update tomorrow.
Thanks :+1:

PR #26, opened 2015-06-21
Annotated all method parameters and return values with android annotations

porokoro:
To improve auto detection of invalid parameters and unexpected return values I annotated all method parameters and return values with the corresponding android annotations.

This gives you warnings directly in IntelliJ and Android Studio (and Lint I guess) when you try to:
- pass a resource id to ``setBarColor`` or ``setSeparatorColor`` instead of a resolved color
- pass _null_ for _Fragment_ or _Context_ to ``addSlide``

Also you get an informational hint that you don't need to check the result of ``getSlides`` for _null_ because it is never _null_ if you do so.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -24,4 +24,5 @@ android {
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
     compile 'com.android.support:support-v4:22.2.0'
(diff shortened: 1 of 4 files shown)


PaoloRotolo:
Thanks again @porokoro :)

PR #35, opened 2015-06-28
Allows to set off screen page limit

juankysoriano:
There are a few cases where overriding the default limit for offset screens in ViewPager can be very useful. I personally think that `AppIntro` should let the developer decide whether or not override this value as `ViewPager` does. 

Thanks for this fantastic library btw.


PaoloRotolo:
Looks good, thank you @juankysoriano! :+1:

PR #39, opened 2015-07-03
Support for gamepads

Fleker:
Android TV and gamepad support for moving to the next option and also
exiting the intro


PaoloRotolo:
Looks good, thank you!

Issue #42, opened 2015-07-08
Please Add This Indicator New Wersion


asdstyle33:
Please Add this New version
https://github.com/LyndonChin/AndroidRubberIndicator

porokoro:
At the moment of writing the linked project seems to be in a development status. To be more precise:
- It seems not to be published to maven, jcenter or other repositories
- The source code looks like there are plans to make it more customizable but at the moment it is very limited in customization options (no color can be set to match the AppIntro design)

This would require to include the whole source of the library repository directly into AppIntro. And that's what I really would like to avoid.

But Based on the interest of other users in this library, perhaps a separate plugin can be provided. This could be separated from the main AppIntro library and would extend the basis with a complete implementation to enable this indicator with a one liner. This way the main library can stay unaffected for users that aren't interested in this additional indicator and the ones who are just need to add the optional dependency and add a one liner. (This solution would require my [pull request](https://github.com/PaoloRotolo/AppIntro/pull/46))

What do you think @PaoloRotolo ?

PaoloRotolo:
@porokoro +1, awesome work!

PR #46, opened 2015-07-14
Custom indicators

porokoro:
At the moment there is only the "Dot"-Indicator. As seen in issue #42 there is the demand to set a custom indicator instead. This pull request abstracts the indicator into an interface to make it able for users to define a **custom indicator**.

For the scenario where there are **a lot of slides** the dots could require more space than available on the device screen. Therefore I added a **progress indicator** for users that want to use a large amount of slides. For example to show a full fledged tutorial.

Perhaps the progress indicator should get a *custom style* to better match the whole design of AppIntro **before the next release** is made.


PaoloRotolo:
Thanks again! Really useful.

PR #51, opened 2015-07-17
Missing dependecy from AppCompat

marbat87:
Sorry :)


PaoloRotolo:
np, looks better now :D

marbat87:
Actually I'm not really good at using Github application :D

PR #54, opened 2015-07-19
Hide the indicator when there is only one slide

porokoro:
At the moment the indicator of the AppIntro is even shown when there is only one slide.
This looks somewhat strange in my opinion because there is no real use to show it for only one slide present.


PaoloRotolo:
LGTM, thanks!

PR #55, opened 2015-07-19
Remove deprecation warning for Resources.getDrawable(int)

porokoro:
The function ```Resources.getDrawable(int)``` is deprecated for newer APIs. In the class **DefaultIndicatorController** there was a method to call the right function already. I moved that function into a separate class to use it in **AppIntroFragment** as well.


PaoloRotolo:
Nice, thanks!

PR #58, opened 2015-07-22
Need a way to get the pager to suppport using CircleIndicator library.

smanikandan14:


smanikandan14:
There should be a way to include CircleIndicator kinda of library to change view pager indicators. For that viewpager needs to be exposed through a getter 

https://github.com/ongakuer/CircleIndicator

```
	class CustomIndicatorController implements IndicatorController {
		
		@Override
		public void initialize(int i) {

		}

		@Override
		public View newInstance(Context context) {
			LinearLayout indicatorLayout = (LinearLayout) View.inflate(context, R.layout.circle_indicator, null);
			CircleIndicator circleIndicator = (CircleIndicator) indicatorLayout.findViewById(R.id.circleIndicator);
			circleIndicator.setViewPager(mPager);
			return indicatorLayout;
		}

		@Override
		public void selectPosition(int i) {

		}
	}
```

PaoloRotolo:
LGTM. Merging...

PR #64, opened 2015-07-29
Fix for page indicator being off center

deinlandel:
Fix for https://github.com/PaoloRotolo/AppIntro/issues/63
(page indicator being off center)


PaoloRotolo:
Thank you! Merging...

PR #65, opened 2015-07-29
Suppress deprecation warning

porokoro:
At the moment the compiler complains about the call to a deprecated function. This warning is not needed because the calling method was created because of the deprecation and handles the call to the right function.


PaoloRotolo:
LGTM. Thanks!

PR #75, opened 2015-08-18
add color options for title and desc

salimkayabasi:
need more color options about title and desc when we decieded to use white as bg color


PaoloRotolo:
LGTM. Thanks!

PR #77, opened 2015-08-24
Support for DPAD, based on reported bug for Nexus Player remote

Fleker:
Android TV remotes rely on a different key than a gamepad. This allows either one to work.


PaoloRotolo:
LGTM, thank you!

PR #93, opened 2015-10-08
Indicator dots color manipulation 

milosjovac:
Indicator dots color manipulation  via setIndicatorColor(selectedColor, unselectedColor) method.


Andrew-Quebe:
Looks good, thanks!

PR #130, opened 2015-12-19
Call onRequestPermissionsResult() of super class in AppIntro and AppIntro2

NiharG15:
- This fixes an issue where requestPermissions() called from a fragment will not deliver the correct result to the fragment.
 - More information here:
   http://stackoverflow.com/a/33995827/2663152
   http://stackoverflow.com/questions/10564474/wrong-requestcode-in-onactivityresult#comment13675437_10564488 (Works similarly for this)

### DIFF
--- a/library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java
+++ b/library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java
@@ -588,6 +588,7 @@ public abstract class AppIntro extends AppCompatActivity {
 
     public void onRequestPermissionsResult(int requestCode, @NonNull String permissions[],
                                            @NonNull int[] grantResults) {
+    	super.onRequestPermissionsResult(requestCode, permissions, grantResults);
         switch (requestCode) {
             case PERMISSIONS_REQUEST_ALL_PERMISSIONS:
                 pager.setCurrentItem(pager.getCurrentItem() + 1);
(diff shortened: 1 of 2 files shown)


Andrew-Quebe:
I'm not sure why this is necessary...the library has a permission result listener but nothing happens in it on purpose. The UI of the app is not supposed to be locked up if someone chooses not to give an all access to their contacts.

NiharG15:
If you request a permission in a fragment (not using the library's inbuilt functionality), for example on a button press, the result is not correctly delivered back to the fragment as the method does not call super.onRequestPermissionResult(). This happens because the activity changes the request code so that it can know which fragment to deliver the result to. (https://android.googlesource.com/platform/frameworks/support/+/refs/heads/master/v4/java/android/support/v4/app/FragmentActivity.java#820). If the super class method is not called, the result is not delivered to the requesting fragment properly.

Andrew-Quebe:
Ok...but why is this needed if the library has inbuilt permissions support?

NiharG15:
In my use case, for example, the app needs to request permission after a button click in one of the slides, and not when the next button is clicked. 
Regardless of this, not being able to get the result reliably is a bug, in my opinion.

Andrew-Quebe:
@NiharG15 alright...thanks for clarifying :) I can see that this PR doesn't affect the inbuilt permissions support, so I guess it's worthwhile. Thank you!

Also, about the getting permission result, it's not a reliability issue. The library is specifically designed to just introduce and request the permissions your app needs...not lock up the UI.

PR #138, opened 2015-12-31
Add description spannable support

SandroMachado:
This PR adds support to add a slide description with a spannable string.

baconlmy48bsandromachado12312015005934

PR #141, opened 2016-01-02
Use CharSequence instead of String

tasomaniac:
With the usage of `CharSequence` object, the users of the library can send `SpannableString` to adjust the styles of the texts.

### DIFF
--- a/library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java
+++ b/library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java
@@ -332,7 +332,7 @@ public abstract class AppIntro extends AppCompatActivity {
      *
      * @param text your text
      */
-    public void setSkipText(@Nullable final String text) {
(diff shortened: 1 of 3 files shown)


Andrew-Quebe:
#138 is causing conflicts with this PR...please make a new one that keeps the Spannable text feature. Thank you :)

tasomaniac:
I was going to do that but it is almost impossible with #138 merged. I would suggest you to revert that merge. 

It is using `Html.fromHtml` first. And then in `newInstance` method, it uses `Html.toHtml` to revert it and then uses `Html.fromHtml` one more time in `onCreateView`. That is bad for performance and does not make sense. 

Besides, `Html.fromHtml` is not the only way to use Spans.

Andrew-Quebe:
Perhaps you could include a more performance friendly way of using spans?

tasomaniac:
I already sent a PR about it. Can you check #143

PR #143, opened 2016-01-04
Use CharSequence instead of String

tasomaniac:
`AppIntroFragmet` was overkill. Even the example with `Html.fromHtml` works now and it is simpler.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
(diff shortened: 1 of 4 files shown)


SandroMachado:
Why I didn't used the `CharSequence` instead of `Spanned` on my PR!!! :) :+1:

tasomaniac:
@SandroMachado exactly. `CharSequence` is the mother of all. :)

Andrew-Quebe:
@tasomaniac thanks for your work :)

PR #193, opened 2016-03-20
Fix not fully rendered ripple effect

MarekSalat:
The ripple effect on buttons has not been fully visible (see
http://imgur.com/HTJFdFC). The fix solves the issue by adding
clipChildern and clipToPadding attributes to all parents’ layouts (see
http://goo.gl/oqbCiu for more information)

### DIFF
--- a/library/src/main/res/layout-v21/intro_layout.xml
+++ b/library/src/main/res/layout-v21/intro_layout.xml
@@ -2,7 +2,9 @@
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="fill_parent"
     android:layout_height="wrap_content"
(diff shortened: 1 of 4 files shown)


MarekSalat:
I have found out it relates to the https://github.com/PaoloRotolo/AppIntro/issues/116

avluis:
Can you update this PR with latest master?

avluis:
I've been testing the changes in this PR with the current master -- so far the original issue has seen changes from the original bug.
At the same time, I'm not seeing this fix actually address the issue entirely.

Andrew-Quebe:
@MarekSalat what's the status of this? If we don't here back soon, we'll need to close this PR.

PR #202, opened 2016-04-03
Gradle Version: Go back to stable version/Import fix

maxee:
Going back to stable gradle plugin version prevents annoying warnings when using an outdated pre-release.

Inludes #199.


maxee:
This fix is also included in my more complex PR #204

PR #204, opened 2016-04-04
[WIP] Refactoring, bugfixes add slide policy

maxee:
As the maintainers seem quite inactive for a few month now, I decided to create a PR which addresses a lot of issues and implements a few new features, I'd like to have.

@PaoloRotolo and @Andrew-Quebe: If you don't want to merge the whole PR but only parts, let me know, for most points I mention in the following list, it would be possible to create a own PR, if requested.

**Note:** All changes are backward compatible and don't break any implementation.

- [x] ~~Go back to stable version of the gradle plugin to prevent warnings when using an outdated preview version (45e7a1a)~~ Edit: Plugin version changed to v2.0.0 stable (768680e)
- [x] Add missing import (576392e). Same as #199 and #207 
- [x] The classes `AppIntro` and `AppIntro2` share a lot of common code. This is ugly, a common parent class would solve it (7c4008b and c0b2d53)
- [x] The function `onNextPressed` is useless. (Marked as deprecated in 55c7034)
- [x] #134, #175 : Add new functions with changed signature of `onSlideChanged`, `onDonePressed` and `onSkipPressed` returning the instance of the current fragment, mark the old ones as deprecated. (55c7034)
- [x] #129: Add the possibility to have some fragment-based policy on whether ~~going to the next slide, pressing done, pressing skip (#100)~~ leaving the intro by pressing the back button is enabled. E.g.: One has a fragment with a license text which should be acknowledged by ticking a checkbox. The fragment should be able to lock the intro until the user checked the checkbox. (Method `setGoBackLock` implemented d014eed)
- [x] #119: Let users override `onCreate()`, mark `init()` as deprecated (2941c7f)
- [x] remove custom method `restoreLockingState` and override `onRestoreInstanceState` instead (2941c7f).
- [x] Consolidate layout files `layout_intro.xml` and `layout_intro2.xml`. Both have a almost completly indentical file for the layout-v21 stuff. This could be easier done via a small style. Also as mentioned in #142, both contain unnecessary color declarations (2153055 and b9b36c9)
- [x] Implement support for immersive mode. Add two methods `setImmersiveMode(isEnabled)` and `setImmersiveMode(isEnabled, isSticky)`. The former is a shorthand method for the latter with the second parameter set to false. (c8e7825)
- [ ] Implement correct fragment instance saving and restoring (for example on orientation changes). 
[...]

PR #207, opened 2016-04-11
fix: import ImageButton in AppIntro class

PedramVeisi:


maxee:
This is the same as #199 and also included into #204

PR #211, opened 2016-04-16
Skip button for Intro2

timobaehr:
skipbuttonintrotour-landscape
skipbuttonintrotour

### DIFF
--- a/library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java
+++ b/library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java
@@ -1,14 +1,9 @@
 package com.github.paolorotolo.appintro;
 
 import android.graphics.drawable.Drawable;
-import android.os.Bundle;
 import android.support.annotation.ColorInt;
 import android.support.annotation.DrawableRes;
-import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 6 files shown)


Andrew-Quebe:
Upon reviewing your code, it seems that a large amount of code has been removed from the `AppIntro` class. Any particular reason why?

library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java
```diff
@@ -17,78 +12,11 @@
 public abstract class AppIntro extends AppIntroBase {
     private static final String TAG = "AppIntro";
 
-    protected boolean skipButtonEnabled = true;
```
Andrew-Quebe: starting here, why was all this removed?

timobaehr: Hi, short answer is "moved" not "removed". 

In the current master the skip button is only used inside the ```AppIntro``` but not in ```AppIntro2```. I moved the skip button code from ```AppIntro``` into the ```AppIntroBase``` class as the skip button could be part of both intro tour variants.

Andrew-Quebe: Ah, very well.

PR #241, opened 2016-06-02
Set visibility of Skip button

RafaG:
If you want old style AppIntro2 without Skip button, you can set setSkipButtonEnabled(false).

### DIFF
--- a/library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java
+++ b/library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java
@@ -100,17 +100,6 @@ public abstract class AppIntro extends AppIntroBase {
     }
 
     /**
-     * Setting to to display or hide the Skip button. This is a static setting and
-     * button state is maintained across slides until explicitly changed.
-     *
-     * @param showButton Set true to display. False to hide.
-     */
-    public void showSkipButton(boolean showButton) {
-        this.skipButtonEnabled = showButton;
-        setButtonState(skipButton, showButton);
-    }
-
-    /**
      * Shows or hides Done button, replaced with setProgressButtonEnabled
      *
      * @deprecated use {@link #setProgressButtonEnabled(boolean)} instead.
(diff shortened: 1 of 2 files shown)


RafaG:
Missing setter.

avluis:
Why not move the method in AppIntro to AppIntroBase so that they are shared across AppIntro and AppIntro2?

RafaG:
I think that functionality of Skip button is well controlled by skipButtonEnabled in AppIntroBase and we have a method to see the status, but not to changed it.

You need to change skipButtonEnabled to false in order to work correclty setProgressButtonEnabled().

avluis:
I don't think you understand what I meant - I'm not talking about why adding a way to disable the skip button in AppIntro2 - that is fine; I am talking about adding an additional method to the library, which is what this PR does.

If you move the following from AppIntro into AppIntroBase, would this not do the same as this PR?
```java
public void showSkipButton(boolean showButton) {
    this.skipButtonEnabled = showButton;
    setButtonState(skipButton, showButton);
}
```

RafaG:
Ok, now I understand you, I was working only with AppIntro2 and don't saw this method .For sure is better what you think.

avluis:
Awesome, if you update the PR with what I have commented above AND remove that same code from AppIntro, I will bring it up with @Andrew-Quebe to see if he agrees on the merge.

avluis:
Better, but that same method has to be removed from AppIntro as well, go ahead and get that change in there as well.
Should be from [here](https://github.com/PaoloRotolo/AppIntro/blob/master/library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java#L102) to [there](https://github.com/PaoloRotolo/AppIntro/blob/master/library/src/main/java/com/github/paolorotolo/appintro/AppIntro.java#L111).

avluis:
Awesome, thanks again~

RafaG:
Thanks a lot. I never worked with git before and I was not so sure what I was doing. :-)

avluis:
@PaoloRotolo or @Andrew-Quebe Do you mind taking a look at this PR and merge if it satisfies you? :eyes:

PaoloRotolo:
Sure, looks good to me. Sorry for the delay.

PR #244, opened 2016-06-07
Use @colorInt annotation

laurentsebag:
The first time I used the library, I used colors like `R.color.primary` by mistake and the intro screen didn't behave as I expected as it expects a color integer 0-255.

Adding the @colorInt annonation warns users doing the same mistake directly in Android Studio.


avluis:
It's a pretty good thing to have annotations throughout a library - specially one that is not documented as well in the code itself - this looks good to me.

PR #271, opened 2016-08-07
Fixed crashing when passing XML vector drawable as image.

xsorifc28:
Sets the image resource rather than the image drawable.

Also updated to latest support libs and build tools


avluis:
Please remove the update to the build tools -- there are already PRs in line to address that.
Other than that, have you tested this PR with the sample app, meaning that non-vector work with no issue, and that nothing is broken?

xsorifc28:
Tested with API 24, (Nougat, 7.0) and API 18 (Jelly Bean, 4.3) using an XML and PNG drawable. Both work fine. 

I don't see this breaking anything, since it uses base ImageView method `setImageResource`. Should work fine across all APIs that support this method.

avluis:
And from my own testing, the example app has no issue with this change.
So I'm merging, thanks for the fix.

PR #291, opened 2016-08-30
German translations

K0bin:
I translated the two strings to German.

### DIFF
--- a/library/src/main/res/values-de/strings.xml
+++ b/library/src/main/res/values-de/strings.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="skip_button">ÜBERSPRINGEN</string>
+    <string name="done_button">FERTIG</string>
+</resources>


avluis:
Neat -- thank you~

Andrew-Quebe:
Merging this now, but there may also be programmatically generated text that will need translating as well. Most of that, however, can be translated on the user side when they use the library.

Andrew-Quebe:
@K0bin refer to issue #5 again if you want to help translate the example app strings.

K0bin:
I'm going to do it but to be honest I don't see the point in translating the example app.

Andrew-Quebe:
@K0bin it's not mandatory, so if you don't want to, that's fine.

library/src/main/res/values-de/strings.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="skip_button">ÜBERSPRINGEN</string>
```
heinrichreimer: Using all-caps text in string resources is an anti-pattern.
Should probably be handled by the `TextView`.

PR #296, opened 2016-09-14
Fix vector drawables in xml

K0bin:
Using android:src resulted in a crash on pre lollipop devices.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -13,6 +13,7 @@ android {
         versionCode 12
         versionName "4.0.0"
         consumerProguardFiles 'proguard-project.txt'
+        vectorDrawables.useSupportLibrary = true
     }
 }
 
--- a/library/src/main/res/layout/intro_layout.xml
+++ b/library/src/main/res/layout/intro_layout.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="fill_parent"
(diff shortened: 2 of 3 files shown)


avluis:
Hmm -- I'd like to hold off this PR for a bit -- making use of `vectorDrawables.useSupportLibrary = true` has a lot of side effects plenty of novice programmers have not dealt with just yet (I know this because I am using it on my production app).

Is there any particular reason why the library needs to have this applied as well?
I am using this library on my production app and it works on pre-lollipop devices without the change here.

avluis:
@K0bin OK, just got done testing to check for side-effects in the library when using this flag.

Tested on:
Moto E (Android 4.4.4)
Lenovo Tab2 (Android 4.4.4)
Moto G3 (Android 5.1.1)
LG G4 (Android 6.0)
LG G Pad X 10.1 (Android 6.0)
Nexus 6 (Android 6.0.1)
Nexus 6P (Android 7)

Everything is working as it is supposed to, therefore, good to merge~

K0bin:
I'm using the flag in production as well and haven't noticed any side effects but that got me curious now.

avluis:
@K0bin 
I'm mostly referring to instances where a compound drawable is in use, and it happens to be a vector drawable.
In most cases, this needs to be used if you are setting your drawables dynamically:
```java
static {
        AppCompatDelegate.setCompatVectorFromResourcesEnabled(true);
    }
```
In the case of compound drawables, you'll need to make use of:
```java
textView.setCompoundDrawablesWithIntrinsicBounds(R.drawable.my_vector_drawable, 0, 0, 0);
// or
textView.setCompoundDrawables(R.drawable.my_vector_drawable, 0, 0, 0);
```
And finally, in the case of notifications (for 4.4.4 and below), you'll need to reference PNGs and not Vector Drawables.

PR #297, opened 2016-09-15
Update README.md

friederbluemle:
Just a few minor README.md updates. It was still referring to `init()` in a few places - which was replaced with `onCreate` in the latest version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,29 +22,28 @@ repositories {
 }
 
 dependencies {
-  compile 'com.github.paolorotolo:appintro:4.0.0'
+    compile 'com.github.paolorotolo:appintro:4.0.0'
 }
 ```
 
 Create a new **Activity that extends AppIntro**:
 
 ```java
-public class MyIntro extends AppIntro {
-
+public class IntroActivity extends AppIntro {
     @Override
-    protected void onCreate(Bundle savedInstanceState) {
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
-        // Add your slide's fragments here.
+        // Add your slide fragments here.
         // AppIntro will automatically generate the dots indicator and buttons.
-        addSlide(first_fragment);
-        addSlide(second_fragment);
-        addSlide(third_fragment);
-        addSlide(fourth_fragment);
+        addSlide(firstFragment);
+        addSlide(secondFragment);
+        addSlide(thirdFragment);
+        addSlide(fourthFragment);
 
         // Instead of fragments, you can also use our default slide
         // Just set a title, description, background and image. AppIntro will do the rest.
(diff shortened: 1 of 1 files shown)


avluis:
Plenty of the edits are good, but why this change:
```java
[...] to //...
```
Making it easier for copying and pasting?

README.md
```diff
@@ -114,16 +113,20 @@ There's no need to create one class for fragment anymore. :)
 No problem, just use this method and AppIntro will generate a new slide for you.
 
 ```java
-addSlide(AppIntroFragment.newInstance(title, description, image, background_colour));
+addSlide(AppIntroFragment.newInstance(title, description, image, backgroundColor));
```
avluis: Grap this bit in onCreate as well.

README.md
```diff
@@ -22,29 +22,28 @@ repositories {
...
 
 Create a new **Activity that extends AppIntro**:
 
 ```java
-public class MyIntro extends AppIntro {
-
+public class IntroActivity extends AppIntro {
     @Override
-    protected void onCreate(Bundle savedInstanceState) {
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
```
avluis: Don't annotate this just yet -- need to make a quick change on the library first.

friederbluemle: It looks like `AppIntroBase` already has this annotation: https://github.com/PaoloRotolo/AppIntro/blob/master/library/src/main/java/com/github/paolorotolo/appintro/AppIntroBase.java#L75

README.md
friederbluemle: Not sure what you mean by that..

friederbluemle:
> @friederbluemle 
> Plenty of the edits are good, but why this change:
> 
> [...] to //...
> Making it easier for copying and pasting?

Hah, that's just a minor change because `[...]` is not "valid" Java code. Anyway, I can revert if you want.

README.md
avluis: Go ahead and do the same for `AppIntro`.

avluis:
> > @friederbluemle 
> > Plenty of the edits are good, but why this change:
> > 
> > [...] to //...
> > Making it easier for copying and pasting?
> 
> Hah, that's just a minor change because `[...]` is not "valid" Java code. Anyway, I can revert if you want.

Nah, that's ok -- I prefer it this way as well -- proper code and less mistakes for others to make.

README.md
avluis: Ah, good then, go ahead with the change.

README.md
friederbluemle: Oh, got it! Sure, good suggestion :+1: I wrapped it as well.

avluis:
I see the changes -- thank you for taking the time to contribute~

PR #315, opened 2016-09-29
Possibility to load image from web + long description texts are now scrollable

siberianbearr:
When creating a new AppIntroFragment, you can pass an URL instead of a image drawable id

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -20,4 +20,5 @@ android {
 dependencies {
     compile 'com.android.support:appcompat-v7:24.2.1'
     compile 'com.android.support:support-annotations:24.2.1'
+    compile 'com.squareup.picasso:picasso:2.5.2'
 }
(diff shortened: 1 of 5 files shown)


avluis:
Two things with this PR I don't agree with:
- No need for AppIntro to have an additional dependency.
Load your images, then feed it to AppIntro.
- If your text needs to scroll, you have way too much text or you need to create your own layout, not make it scrollable.

@Andrew-Quebe I'm leaving this one up to you.

Andrew-Quebe:
I agree with @avluis. If a user wants to load their images in from URLs, they can do so and then pass that to the library rather than the library having unnecessary dependencies for the folks that don't with to load their images this way.

Scrollable text is a total UI no-no. As stated above, if you have so much text that it needs to fit into a scrollview, you're doing your text wrong. An app intro is meant to be a subtle, brief introduction to what your app does. If you introduce your users to your app and it takes you a paragraph to do so, you're doing something wrong. AppIntro is not meant for that, and therefore will not be supporting the ability to do so.

PR #353, opened 2016-11-14
added turkish language support

livet96:


avluis:
You only have to translate strings for the library (not the same app)~

livet96:
Oh okay.I'll check tomorrow.

Andrew-Quebe:
Reopen when you've only changed the two strings in the library.

PR #365, opened 2016-11-28
Added arabic strings

smidaharoun:

PR #366, opened 2016-11-28
removed example app translate and improve

livet96:


avluis:
As previously mentioned -- the only file that you need to worry about translating is the library's single [string.xml](https://github.com/paolorotolo/AppIntro/blob/master/library/src/main/res/values/strings.xml) file -- for example, check PR #365

Andrew-Quebe:
@livet96 thanks!

Andrew-Quebe:
@livet96 refer to issue #5 again if you want to help translate the example app strings.

PR #415, opened 2017-02-19
Allow users to change Bar color on AppIntro2, too

almic:
I think this will be a nice addition 💯 

As far as I can see, this will definitely work since the ID already exists in intro_layout2


paolorotolo:
LGTM, thanks!

PR #420, opened 2017-03-06
Add 'layout_above' attribute to anchor ViewPager fragments.

rcraig-quarksecurity:
Avoid overlapping views in cases where the added
Fragments merge with the bottom layout.

### DIFF
--- a/library/src/main/res/layout/intro_layout.xml
+++ b/library/src/main/res/layout/intro_layout.xml
@@ -9,7 +9,8 @@
     <com.github.paolorotolo.appintro.AppIntroViewPager
         android:id="@+id/view_pager"
         android:layout_width="fill_parent"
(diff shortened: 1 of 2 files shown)


rcraig-quarksecurity:
This patch would unfortunately prevent each slide from setting a different background color in a uniform way. Not having the viewpage anchored seems to allow its background color to bleed over into the navigation portion and then switch to the next prescribed color when switching to the next slide (by the looks of it the intended purpose of the library).

My issue was with a slide that held too much content and therefore ran over top the navigation elements. I guess by having too much content in one slide it breaks with your library's intended use as a simple set of intro slides.

I'll close this myself. Sorry for the noise.

avluis:
Thanks for giving this a try and thanks a lot for taking the time to test it so thoroughly.
It's very appreciated.

PR #459, opened 2017-07-06
Set Random Animation

pacdouglas:
I was in such doubt of which animation to choose that I created this option to randomly choose it. Lol. Thank you very much for your lib, it was very usefull!

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,8 @@ setZoomAnimation()
 setFlowAnimation()
 setSlideOverAnimation()
 setDepthAnimation()
+
+setRandomDefaultAnimations() - Draw one of the above options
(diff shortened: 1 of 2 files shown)


paolorotolo:
Hey, thanks for your pull!
The method works but I don't think it should be in our library, the usage would be very limited.

@Andrew-Quebe @avluis what is your opinion on this?

avluis:
@paolorotolo I like that initiative was taken to share modifications to the codebase, but I do agree -- the use case for this would be very limited.

paolorotolo:
Okay! @pacdouglas thanks again, but I'm closing the pull request.

PR #488, opened 2017-10-06
Feature: replace fill_parent with match_parent

### LINKED ISSUE #481: Constant fill_parent is deprecated
I see in your librarry, you use `fill_parent` in `res/layout..` the best way is set `match_parent`

https://stackoverflow.com/questions/5761960/what-is-the-difference-between-match-parent-and-fill-parent

rberghegger:
This pull request closes #481 by replacing all occurrences of "fill_parent" with "match_parent".

Since the specified minimum SDK version is 14 (for both the example and the library module) it's safe to do that because it's higher than SDK version 8 where "match_parent" was introduced.


avluis:
Thank you for taking the time to make those changes~

PR #493, opened 2017-10-14
Update AppIntroViewPager.java

idish:
There was some logic bug in which caused `isPolicyRespected `of the policy fragments to be called every touch event (and not every "swipe event" as it should) that the ViewPager captures (!). 
That caused some unpredictable behavior in my fragments and morever, redudndant calls to methods.
This PR should be merged ASAP.

Thank you.


avluis:
Can you go a bit more into detail as to those unexpected behaviors -- just trying to get that documented before merging this (thank you for taking the time to create a PR)~

idish:
Closing this PR as it is no longer relevant, I submitted another one.

PR #495, opened 2017-10-18
Update AppIntroViewPager.java

idish:
This PR fixes the "swiping back" issue that many users have complained about.
Users should now be able to swipe to previous pages even when `isPolicyRespected `returns false.

Moreover, `isPolicyRespcted` is now not called every "touch", but every "swipe" detected as it should.

### DIFF
--- a/library/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.java
+++ b/library/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.java
@@ -3,6 +3,7 @@ package com.github.paolorotolo.appintro;
 import android.content.Context;
 import android.support.v4.view.ViewPager;
 import android.util.AttributeSet;
+import android.util.Log;
 import android.view.MotionEvent;
 import android.view.animation.Interpolator;
 
@@ -91,8 +92,7 @@ public final class AppIntroViewPager extends ViewPager {
         if (event.getAction() == MotionEvent.ACTION_DOWN) {
             currentTouchDownX = event.getX();
             return super.onInterceptTouchEvent(event);
-        } else if (checkPagingState(event) || checkCanRequestNextPage(event)) {
-            // Call callback method if threshold has been reached
+        } else if (checkPagingState(event) && checkCanRequestNextPage(event)) {
             checkIllegallyRequestedNextPage(event);
             return false;
         }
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.java
```diff
@@ -3,6 +3,7 @@
 import android.content.Context;
 import android.support.v4.view.ViewPager;
 import android.util.AttributeSet;
+import android.util.Log;
```
avluis: Can you remove unneeded imports from this file (`Ctrl + Alt + L`).

library/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.java
```diff
@@ -177,13 +177,13 @@ private void initViewPagerScroller() {
...
     private boolean detectSwipeToEnd(MotionEvent event) {
-        final int SWIPE_THRESHOLD = 0; // detect swipe
+        final int SWIPE_THRESHOLD = 50; // detect swipe
         boolean result = false;
 
         try {
             float diffX = event.getX() - currentTouchDownX;
             if (Math.abs(diffX) > SWIPE_THRESHOLD) {
-                if (diffX < 0) {
+                if (diffX < 0) {
```
avluis: Please remove the extra spaces here.

library/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.java
```diff
@@ -121,7 +121,7 @@ private boolean checkPagingState(MotionEvent event) {
             return true;
         }
 
-        if (!nextPagingEnabled) {
+        if (nextPagingEnabled) {
```
avluis: I take it this 'fixes' being able to scroll back when making use of `isPolicyRespected` -- but what are the side-effects of changing this eval.
Have you tested swiping without making use of `isPolicyRespected`?

library/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.java
```diff
@@ -140,7 +140,7 @@ private boolean checkCanRequestNextPage(MotionEvent event) {
     }
 
     private void checkIllegallyRequestedNextPage(MotionEvent event) {
-        int swipeThreshold = 25;
+        int swipeThreshold = 50;
```
avluis: Go ahead and define this as a constant to this file -- make use of that here and on `detectSwipeToEnd()`

paolorotolo:
Closing this for inactivity.

PR #519, opened 2018-02-21
Improve RTL checks

efimerdlerkravitz:
Unfortunately I can't squash the request, if you'll decide to accept the request, please squash it.

The current implementation that checks whether it's an RTL device has one specific flaw, it does not check if the developer defined specifically in AndroidManifest `android:supportsRtl="false"`, a usecase might be that an app does not want to support RTL because it does not look good when this flag is toggled on.

I made 2 main changes:
* Removed from the manifest `android:supportsRtl="true"` which is wrong for a library
* Changed `isRtl` from `LayoutUtils` to read what is the configuration the manifest


paolorotolo:
Looks good to me, thanks!

PR #521, opened 2018-02-21
add builder for SliderPage model

Mikelantonio:
Hi,

I developed a builder class to easy create Slider Page.
I add JUnit Test too.


paolorotolo:
Looks good, thanks Mike <3

PR #523, opened 2018-03-04
Add Chinese translations

dlackty:
As titled, both Traditional & Simplified Chinese strings were added.


paolorotolo:
Looks good, thanks!

PR #524, opened 2018-03-15
Correct license name

dlackty:
Usually "Apache License" means version 1, and according to other files, this library actually uses version 2.


paolorotolo:
Looks good, thanks.

PR #541, opened 2018-06-13
Fix RTL support in Wizard Mode

hichamboushaba:
RTL support has some issues when Wizard Mode is enabled, the backbutton is inversed, and the click handler push forward instead of back.


paolorotolo:
LGTM, thanks.

PR #546, opened 2018-07-05
Cleanup some orphan tests file

cortinico:
There are a lot of orphan test files and I'm cleaning up some of them.

Summary:
* removed instrumentation test dependencies from example test app
* Removed not referenced `androidTest` util classes from example test app
* Removed useless `ApplicationTest` from library


cortinico:
Yeah makes sense, I'll prepare a new PR

paolorotolo:
Thanks!

Issue #547, opened 2018-07-08
Bug in isRtl method.


isabsent:
You have to use this definition for **RTL**:

    public static boolean isRtl() {//https://stackoverflow.com/a/37101995/753575
        Locale locale = Locale.getDefault();
        final int directionality = Character.getDirectionality(locale.getDisplayName().charAt(0));
        return directionality == Character.DIRECTIONALITY_RIGHT_TO_LEFT ||
                directionality == Character.DIRECTIONALITY_RIGHT_TO_LEFT_ARABIC;
    }
instead of your because `android:supportsRtl="true"` in the manifest does not mean user's device runs in **RTL** mode. It only means you are supporting **RTL** in your app.

Anutha02:
@cortinico Our library correctly detects RTL. This issue is invalid. Please close

AnuthaDev:
@cortinico Please close this

Issue #545, opened 2018-07-05
Readme Sample is partially deprecated


cortinico:
I was trying the library and I noticed that the first example in the `README.md` file contains deprecated code.

I'm talking about this section:

```java
public class IntroActivity extends AppIntro {
    @Override
    protected void onCreate(@Nullable Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

[...]

        // Instead of fragments, you can also use our default slide
        // Just set a title, description, background and image. AppIntro will do the rest.
        addSlide(AppIntroFragment.newInstance(title, description, image, backgroundColor));
```

The `AppIntroFragment.newInstance` method is deprecated [here](https://github.com/apl-devs/AppIntro/blob/master/library/src/main/java/com/github/paolorotolo/appintro/AppIntroFragment.java#L14)
Maybe that should be replaced with the `newInstance(SliderPage sliderPage)` variant?

AmiroEmad:
I have same issue what should I do to be able to use the library ??

AmiroEmad:
Thank you . Your issue helped me . I went and checked the code and found the solution to my problem . Thank you so much . @cortinico

cortinico:
You're welcome :)

cortinico:
Closing as this is going in 6.0.0

PR #548, opened 2018-07-12
Updating README.md to do not use deprecate methods (#547)

cortinico:
I just replaced the old `newInstance` in the README with the new one.


cortinico:
Closing since I referenced the wrong Issue (#547 instead of #545)

PR #549, opened 2018-07-12
Updating README.md to do not use deprecate methods (#545)

cortinico:
I just updated the README to use the non deprecated `newInstance`


gcantoni:
+1

PR #554, opened 2018-07-30
Switch 'compile' to 'implementation' statements in gradle script of the library

idish:
Updated build.gradle script to use the new 'implementation' syntax instead of the old 'compile'


cortinico:
This CR can be discarded as we already updated to `compileSDK` 28 and migrated all the dependencies to `implementation`

PR #564, opened 2018-08-20
Converted SliderPage to a Kotlin Data class

cortinico:
I converted `SliderPage` to a data class, just because was a lot of boilerplate.
Was wondering if we want to have the `@JvmOverload` to generate multiple constructors or force to provide all the parameters (IMHO we should be fine with the JvmOverload)


paolorotolo:
Yes, I think we should be fine with JvmOverload. We also had an empty constructor in the original class so I don't see why we should force users to provide all parameters when the object is created now. :)

paolorotolo:
LGTM, merging

Issue #565, opened 2018-08-22
Next and Done buttons need accessibility descriptions


jrcacd:
Google Play Console complains about this

cortinico:
@jrcacd The warning should go away after my PR gets merged.

paolorotolo:
Fixed with #566

PR #566, opened 2018-08-22
Added missing accessibility labels for all the buttons #565

cortinico:
Fix for #565 

Unfortunately we need to translate two more strings to all the supported languages. Don't really know how we do that? Should we just fallback to `en` for accessibility?


paolorotolo:
Hey @cortinico! Thanks for the fix.

I think we can use ```en``` for now and hope that other contributors will add translations later.

PR #572, opened 2018-08-29
Fix for titleString & descriptionString in the SliderPage

cortinico:
Looks like a broke two fields of `SliderPage` when I converted it to Kotlin 😅

both `titleString` and `descriptionString` should not have a backing field.
We want those properties to behave as getters and do not have an associated value.

The current implementation was causing several titles to disappear:
screenshot_1535578033


paolorotolo:
Okay, thanks!

Issue #573, opened 2018-09-04
Add suffix or prefix to resource names


eGorets:
`fragment_intro` is pretty common name for projects and now I have a conflicts. I think it's a good practice to separate resources by its namespace.

paolorotolo:
Good point, thanks for reporting this.

paolorotolo:
Closed with #574.
Thanks.

PR #574, opened 2018-09-05
Add prefix for resources (Closes: #573)

paolorotolo:


paolorotolo:
Also updates gradle to -rc02

Issue #580, opened 2018-09-29
I wonder if the method setAnimationColors(ArrayList<Integer> colors) is useful.


qianshengta:
**AppIntro Version**:
AppIntro Version：4.2.3

**Device/Android Version**:
null

**Issue details / Repro steps / Use case background**: 
I looked at the source code，ArrayList set through "setAnimationColors(ArrayList<Integer> colors)" are not used by any objects.I wonder if the method setAnimationColors(ArrayList<Integer> colors) is useful.

**Your Code**: 
null

**Stack trace / LogCat**:
null

cortinico:
Yeah you're right. That method can be removed.
Here the PR @paolorotolo #582

paolorotolo:
#582 is merged now.
Thanks for the contribution.

PR #582, opened 2018-10-02
Remove unused setAnimationColors (#580)

cortinico:
The method and the backing field are basically unused completely.


paolorotolo:
Thanks! LGTM.

PR #585, opened 2018-10-07
Add Vietnamese translation and upgrade gradle build to 3.2.0, change android to androidx 

chihung93:
Hello, I have done some task in this library:
1. Add Vietnamese translation
2. Upgrade Gradle to 3.2.0 (base on Android studio suggestion)
3. Change Annotation base on Android studio suggestion (android to androidx)

Tks.


example/src/main/java/com/amqtech/opensource/appintroexample/SliderPageModel.java
```diff
@@ -0,0 +1,96 @@
+package com.amqtech.opensource.appintroexample;
+
+import androidx.annotation.ColorInt;
+import androidx.annotation.DrawableRes;
+
+/**
+ * Created by tatianasolonets on 3/19/17.
+ */
+
+public class SliderPageModel {
```
cortinico: Any reason why this file was re-added?

PR #588, opened 2018-10-11
Cleanup some orphan tests file - round 2

cortinico:
This is a follow-up for #546 

Body from previous PR:

There are a lot of orphan test files and I'm cleaning up some of them.

Summary:

* removed instrumentation test dependencies from example test app
* Removed not referenced androidTest util classes from example test app
* Removed useless ApplicationTest from library


paolorotolo:
Nice work. And yes, we need to write proper tests. We'll talk about this on #589

PR #590, opened 2018-10-13
Using Fonts from Resources

Javinator9889:
At this Pull Request, I modified and included the code for using **custom Fonts** that are available at the `res/fonts` by providing a *FontId*: `R.font.raleway`, using new **Android X** libs and compatibility by using `ResourcesCompat`.

It uses the same class that was working before (`CustomFontCache`) with new class for managing fonts and make it easier (and cleaner) defining new fonts: `TypefaceWorker`.

With this upgrade in use, the developer **will be able** to include and define fonts also by adding the `FontRes` to any `newInstance` constructor (overloaded constructors with the same arguments but one is `String`, other is `@FontRes int`).

Tested and optimized for working as faster as possible. Also included *JavaDocs* if any developer wants to inherit from class and use the new `TypefaceWorker`


Javinator9889:
@paolorotolo if you finally merge Pull Request I can edit the wiki and add the new methods/necessary entries 😄

library/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.github.paolorotolo.appintro">
 
-    <application ></application>
+    <uses-permission android:name="android.permission.VIBRATE" />
```
cortinico: AFAIK This was intentional. We don't want to propage the `android.permission.VIBRATE` to all the apps that are using this library. If you (android dev) want to use the Vibration you need to define the permission in the app manifest and enable the feature.

library/src/main/java/com/github/paolorotolo/appintro/AppIntro2Fragment.java
```diff
@@ -36,6 +75,20 @@ public static AppIntroFragment newInstance(CharSequence title, @FontRes int titl
...
+     *                      the title found at assets folder
+     * @param description   CharSequence which will be the slide description
+     * @param descTypeface  CharSequence the URL of the custom typeface for
+     *                      the description found at assets folder
+     * @param imageDrawable @DrawableRes (Integer) the image that will be
+     *                      displayed, obtained from Resources
+     * @param bgColor       @ColorInt (Integer) custom background color
+     * @return AppIntro2Fragment created instance
+     */
     public static AppIntroFragment newInstance(CharSequence title, String titleTypeface,
```
cortinico: Ideally this should also be `@deprecated` and we should have a method with a `SliderPage` parameter only (similarly to `AppIntroFragment`).

library/build.gradle
```diff
@@ -10,8 +10,8 @@ android {
     defaultConfig {
         minSdkVersion 14
         targetSdkVersion 28
-        versionCode 14
-        versionName "5.0.0"
+        versionCode 15
+        versionName "5.0.1"
```
cortinico: Since you're doing breaking changes, this is probably going to be a major release or a minor release (not a patch bump)?

Javinator9889: Well, there is not exactly any "breaking changes" as it is the same code with the same Android X libs version, only increased `versionCode` and `versionName`
So I think it is not exactly a "major release" as many changes are just a refactor

library/src/main/java/com/github/paolorotolo/appintro/AppIntro2Fragment.java
Javinator9889: Applied on commit 0b83258

library/build.gradle
cortinico: My point was related to the PR. You're removing this method:

`public void setSkipTextTypeface(@Nullable final String typeURL)`

From the `AppIntro` class right? So whoever is currently using this method on `5.0.0` and will bump to `5.0.1` will not be able to build anymore.

library/src/main/AndroidManifest.xml
Javinator9889: Solved on commit 0ff4cc7

library/build.gradle
Javinator9889: I didn't realized that
Sorry, will recover method again and also included new options to change `skipTextTypeface` by using `@FontRes`

Javinator9889: @cortinico I noticed that I did not remove that method, just change its param from `String` to `Object`, so the user is able to also provide a `@FontRes` 
But perfectly it is possible to define two overloaded methods, where one provides an `String`, other an `Integer`

Javinator9889: @cortinico methods recovered on commit 83c2ff295ab46bce137b8fc5916788b3597a9447

Javinator9889:
Refactoring completed at commit 935a5b7

Javinator9889:
@paolorotolo removed Object when possible and overloaded methods
It is not possible to remove it from SliderPage as there is no possibility to use two constructors in Kotlin

paolorotolo:
@Javinator9889 yes, it does! :)

Read here: http://kotlinlang.org/docs/reference/classes.html#secondary-constructors

Javinator9889:
@paolorotolo check if now it is enough :)
I think there is no more places to replace `Object` where possible

paolorotolo:
Much better now, thanks! About the version number, @cortinico do you think we can jump to 5.1.0 since support for ```FontRes``` can be considered a feature?

Javinator9889:
@paolorotolo done
Now just resolve `build.gradle` conflicts and it is done 👍 :)

cortinico:
@Javinator9889 So `5.1.0` looks fine to me. Anyway the maintainer is @paolorotolo so to him the final call ;)

README.md
```diff
@@ -1,355 +1,355 @@
...
+			...
+			maven { url 'https://jitpack.io' }
+		}
+	}
+```
+
+2. **Add the dependency**
+```gradle
+	dependencies {
+	        implementation 'com.github.paolorotolo:appintro:v5.1.0'
```
paolorotolo: Ok, you probably hate me right now, but just one more thing :D
Can you keep the readme version to 5.0.1 (current release?). I want to add some other things before releasing 5.1.0 to the public.

Javinator9889: @paolorotolo no problem, will do tomorrow ^^
Glad to be helpful :smile:

Javinator9889: Changed on commit 11f3e06 @paolorotolo

library/src/main/java/com/github/paolorotolo/appintro/util/TypefaceWorker.java
```diff
@@ -25,26 +28,39 @@ public TypefaceWorker() {
     /**
      * Direct constructor by giving the typeface
      *
-     * @param typeface <p>must be an <b>Integer</b> or a <b>String</b>, first one a FontResource;
-     *                 second one Font Name</p>
-     * @throws RuntimeException <p>when the provided typeface is not an <b>Integer</b> or a <b>String</b></p>
+     * @param typeface <p>a <b>String</b>, which is an URL to font found at Assets folder
      */
-    public TypefaceWorker(Object typeface) {
+    public TypefaceWorker(String typeface) {
```
Javinator9889: As the constructors have changed, it is not possible to merge pull request without breaking build - working on it...

Javinator9889: Solved on commit 5ae2863

paolorotolo:
Looks good now. Thanks a lot!

PR #594, opened 2018-10-18
Do not expose public methods with `Object` parameters (#590)

cortinico:
Follow up to #590 that exposed some `Object` parameters in the public Api.

* Removed all the `Object` occurrencies and all the `instanceof` calls
* Converted `TypefaceWorker` to Kotlin
* Added an usage example in the Sample App
* Updated tests (and converted to Kotlin as well)


cortinico:
Happy to know that the CI is working 😅

paolorotolo:
Looks good, thanks!

PR #600, opened 2018-10-23
Kotlinize the 'util' package

cortinico:
* Kotlinize all the classes in the `util` package`
* Kotlinized `SliderPagerBuilder`
* Establish the `internal` package. All the classes in this package should not be used outside of the library. The idea behind this is to reduce the API surface of the library and to have a package where we can drop all of our utils/public classes. (Please note that this might be a breaking change, although I don't think it requires a major version bump).
* Added tests for the new classes


cortinico:
Yup we had a duplicate `SliderPageBuilderTest`. Now it should work properly.

cortinico:
Tests failed with:
`kotlin.jvm.KotlinReflectionNotSupportedError at LogHelperTest.kt:16`

I'm going to take care of this.

cortinico:
So now it should be good to go ;)

paolorotolo:
Here you go :)

PR #601, opened 2018-10-26
Kotlinize the PermissionWrapper

cortinico:
I converted the `PermissionObject.java` to Kotlin, renamed it, and moved it inside the `internal` package.


paolorotolo:
LGTM, ty!

PR #602, opened 2018-10-29
Kotlinize the 'indicator' package

cortinico:
* Converted all the indicators to Kotlin
* Renamed `DefaultIndicatorController` to `DotIndicatorController`
* Both controllers now are custom view (less code and cleaner lifecycle)
* Removed inflation of views (this will improve performances of the indicators)
* Added a progress indicator in the sample app


paolorotolo:
Looks good!

PR #603, opened 2018-10-29
Convert all the interfaces to Kotlin

cortinico:


cortinico:
Sorry, pushed the wrong commit. Going to follow-up here

PR #605, opened 2018-10-29
Kotlinize the ScrollerCustomDuration

cortinico:
Converted the `Scroller`.

Although @paolorotolo I wasn't really able to make it work (haven't tried much honestly).
Do you know if this is like dead code?


paolorotolo:
Yes, I think it's needed for our public method ```setScrollDurationFactor()``` in AppIntroViewPager. Rewrite looks good BTW, thanks

PR #611, opened 2018-11-06
Kotlinize the ViewPager

cortinico:
Keep on converting to a modern programming language ;)


paolorotolo:
Kotlin is awesome indeed :)

PR #612, opened 2018-11-06
Kotlinize fragments

cortinico:
Please note that this PR contains breaking changes.
I've kotlinized all the `Fragments` in the library.

Moreover, I cleaned up the exposed Api for `AppIntroFragment` and `AppIntro2Fragment`. Now they expose only two methods:

* `newInstance(SliderPage)` as before
* A `@JvmOverload`-ed method like this
```
newInstance(
                title: CharSequence? = null,
                description: CharSequence? = null,
                @FontRes titleTypefaceFontRes: Int = 0,
                @FontRes descTypefaceFontRes: Int = 0,
                @DrawableRes imageDrawable: Int = 0,
                @ColorInt bgColor: Int = 0,
                @ColorInt titleColor: Int = 0,
                @ColorInt descColor: Int = 0
        )
```
This will cover the majority of the use cases. I saw that we had several @deprecated methods, plus other non deprecated `newInstance` methods with a permutation of those parameters.

I generally like the two classes now as they're smaller and cleaner. But I'd love to discuss how much aggressive we want to be with removal of old APIs


cortinico:
ping @paolorotolo 👋

Issue #624, opened 2018-12-31
Missing content description for several UI elements


peterhav:
**AppIntro Version**: v5.1.0

**Device/Android Version**: All

**Issue details / Repro steps / Use case background**:  Most App intro UI elements do not provide a content description. A propor content description is required as described here: https://support.google.com/accessibility/android/answer/7158690). Pre-launch reports indicate these warnings for App Intro section of my app.

**Your Code**:  

Elements such as this one:

```


<ImageButton
                android:id="@+id/next"
                style="@style/AppIntroButtonStyleCompat"
                android:layout_width="wrap_content"
                android:layout_height="match_parent"
                android:layout_gravity="end"
                android:layout_marginEnd="8dp"
                android:layout_marginRight="8dp"
                android:minWidth="100dp"
                app:srcCompat="@drawable/ic_appintro_navigate_next_white"
                tools:ignore="ContentDescription" />
```

Should become:
```
<ImageButton
                android:id="@+id/next"
                style="@style/AppIntroButtonStyleCompat"
                android:layout_width="wrap_content"
                android:layout_height="match_parent"
                android:layout_gravity="end"
                android:layout_marginEnd="8dp"
                android:layout_marginRight="8dp"
                android:minWidth="100dp"
                app:srcCompat="@drawable/ic_appintro_navigate_next_white"
                android:contentDescription="Next"
                tools:ignore="HardcodedText"
               />
```

![image](https://user-images.githubusercontent.com/1724223/50561547-0a97cd00-0d0c-11e9-8a33-5309194ca7a3.png)

cortinico:
Hey @peterhav you're right.
This is a big problem, I'll be fixing it in the upcoming release of the library.

cortinico:
Here the fix #626

cortinico:
So this will be fixed in the upcoming version of the library

PR #626, opened 2019-01-03
Fixing missing Content Description (#624)

cortinico:
I'm updating all the `contentDescription` attributes for the required images/imagebuttons.
Moreover I've added a new string to support the content description of the graphics. This is not translated atm, but will default to _graphics_ for the missing translation.


paolorotolo:
Looks good, thanks!
Also first PR of 2019 \o/

PR #628, opened 2019-01-09
Extracting ViewPager background color to a resource (#335)

cortinico:
The ViewPager background is currently hardcoded inside the layout.
Let's extract it so clients can also override it.


paolorotolo:
Thanks!

PR #629, opened 2019-01-09
Create indonesian translation

moxspoy:


appintro/src/main/res/values-id/string.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <string name="app_intro_skip_button">LEWATI</string>
+    <string name="app_intro_next_button">LANJUT</string>
+    <string name="app_intro_back_button">KEMBALI</string>
+    <string name="app_intro_done_button">SELESAI</string>
```
cortinico: Can I ask you to provide translation also for:

```
<string name="app_intro_image_content_description" tools:ignore="MissingTranslation">graphics</string>
```
https://github.com/paolorotolo/AppIntro/blob/master/appintro/src/main/res/values/strings.xml#L7

Is a string used as content description for the image.

moxspoy: sure. its better if you put that line code in snippet on section 'translating'  in readme

cortinico:
Thanks for the PR, we can merge just after Travis result 👍

paolorotolo:
LGTM, thanks for the contribution

PR #633, opened 2019-01-12
Updating Gradle to 5.1.1

cortinico:
I don't expect major breaking changes :) Let's see what Travis has to say.


paolorotolo:
Yep, it didn't complain much :)

PR #634, opened 2019-01-12
Kotlinize the AppIntroViewPager

cortinico:
Completely rewritten the `AppIntroViewPager` in Kotlin.

I've also cleaned up the touch event management code. I have the feeling previously was not really working. Now I've implemented everything in the `onTouchEvent`. Previously having also the `onInterceptTouchEvent` in place, prevented custom layout widget (like a button) from receiving touch events.


paolorotolo:
Looks good now!

PR #637, opened 2019-02-08
Added PT and changed PT-BR

GuilhE:
- Changed values from values-pt to values-pt-rBR
- Created Portuguese strings.xml in values-pt


appintro/src/main/res/values-pt-rBR/strings.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="app_intro_skip_button">PULAR</string>
```
cortinico: Hey @GuilhE 
I assume you skipped the other strings as they're the same in portuguese and `pt-rBR` right?

GuilhE: Yup exactly :)

PR #642, opened 2019-02-23
Replace the Layout2 background color with the proper resource

cortinico:
Looks like this was left behind when working on #335

The background should be a resource so that developers can override it
eventually


paolorotolo:
LGTM, thanks!

PR #645, opened 2019-02-24
Less hardcoded stuff

gcantoni:


appintro/src/main/res/layout/appintro_intro_layout2.xml
```diff
@@ -31,7 +31,7 @@
             android:id="@+id/bottom_separator"
             android:layout_width="match_parent"
             android:layout_height="1px"
-            android:background="#55000000"
+            android:background="@color/textBg"
```
cortinico: So this `<TextView>` can be removed entirely in this layout file.
Looks like the `bottom_separator` is used only on `layout` and not in `layout2`.
Can I ask you to apply this fix on the other layout file: `appintro/src/main/res/layout/appintro_intro_layout.xml`?

appintro/src/main/res/values/colors.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
     <color name="appintro_default_background_color">#222222</color>
+    <color name="textBg">#55000000</color>
```
cortinico: Please rename to `appintro_default_separator_color` for consistency.

appintro/src/main/res/layout/appintro_intro_layout2.xml
gcantoni: What do you mean with applying this fix to the other layout file ?

Should i make non-hardcoded the color there too or delete the TextView?

cortinico: > What do you mean with applying this fix to the other layout file ?

Sorry so, I'd:
* Remove the `<TextView>` from the `_layout2.xml` file
* Apply the fix (don't hardcode the color) from the `_layout.xml`

Anutha02:
The issue this pull request solves is solved in #647 along with a lot others... So this PR is redundant...

PR #639, opened 2019-02-22
Update Skip Icon and add Hindi translation.

Anutha02:
Updated skip icon to be more consistent with material design and added Hindi Translation.

PR #646, opened 2019-02-24
Reverting 'Update Skip Icon' as the icon is not centered

cortinico:
Reverting part of #639 of @Anutha02 as the Skip icon is not centered in the icon.

screenshot_1551047270

Moreover I think that having two "arrows" for the user is confusing to understand exactly what are doing.


Anutha02:
Ummm, it can be easily centred: Delete the icon, right click drawable folder, add vector asset and choose this icon with the required size. I am currently working on a PR that modifies the bottom bar of layout2 a bit so please wait for that...

paolorotolo:
This is probably fixed by #647. I'm testing it right now.

Anutha02:
Yup. Fixed by #647

PR #660, opened 2019-03-18
[5] Minor Fixes

AnuthaDev:
UI fixes, gradle version update, material drawer library update, consumer-proguard file update

### DIFF
--- a/appintro/src/main/java/com/github/paolorotolo/appintro/internal/viewpager/ViewPagerTransformer.kt
+++ b/appintro/src/main/java/com/github/paolorotolo/appintro/internal/viewpager/ViewPagerTransformer.kt
@@ -65,7 +65,7 @@ internal class ViewPagerTransformer(
                     page.isClickable = true
                 } else {
                     // position is between -1.0F & 0.0F OR 0.0F & 1.0F
-                    page.translationX = page.width * -position
+                    page.translationX = page.width* -position
                     page.alpha = 1.0f - Math.abs(position)
                 }
             }
--- a/appintro/src/main/java/com/github/paolorotolo/appintro/internal/viewpager/ViewPagerTransformer.kt
+++ b/appintro/src/main/java/com/github/paolorotolo/appintro/internal/viewpager/ViewPagerTransformer.kt
@@ -56,16 +56,13 @@ internal class ViewPagerTransformer(
             }
             TransformType.FADE -> {
                 if (position <= -1.0f || position >= 1.0f) {
-                    page.translationX = page.width.toFloat()
                     page.alpha = 0.0f
                     page.isClickable = false
                 } else if (position == 0.0f) {
-                    page.translationX = 0.0f
                     page.alpha = 1.0f
                     page.isClickable = true
                 } else {
                     // position is between -1.0F & 0.0F OR 0.0F & 1.0F
-                    page.translationX = page.width* -position
                     page.alpha = 1.0f - Math.abs(position)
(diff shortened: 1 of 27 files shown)


appintro/src/main/res/values/dimen.xml
```diff
@@ -15,7 +15,8 @@
     <dimen name="appintro_button_minwidth">48dp</dimen>
 
     <dimen name="appintro_indicator_inset">3dp</dimen>
-    <dimen name="appintro_indicator_size">10dp</dimen>
+    <dimen name="appintro_unselected_indicator_size">10dp</dimen>
```
cortinico: Please keep those two together

appintro/src/main/res/values/styles.xml
```diff
@@ -3,22 +3,42 @@
 
     <style name="AppIntroStyle" parent="Theme.AppCompat.DayNight.NoActionBar" />
```
cortinico: I think we should rename this to be `AppIntroTheme` as it's actually a theme and not a style.
Moreover we can rename the Themes in the `example` module to avoid naming clashes

appintro/src/main/res/values/dimen.xml
Anutha02: What does together mean?

appintro/src/main/res/values/styles.xml
Anutha02: Okay

appintro/src/main/res/values/dimen.xml
cortinico: Just have a single:
```
    <dimen name="appintro_indicator_size">10dp</dimen>
```

And not two `appintro_unselected_indicator_size` and `appintro_selected_indicator_size`

Anutha02: No. Disagreed.

cortinico: > No. Disagreed.

Do you have arguments?

Anutha02: Having 2 separate dimensions is for changing the size of the selected indicator dot. We can make it bigger if we want this way

Anutha02: Here are the arguments ^

Anutha02: Something just like this:. 
https://drive.google.com/file/d/18RsCpEWXe8CCYqu3gNcVjfVdeAB1tkOA/view?usp=drivesdk

appintro/src/main/java/com/github/paolorotolo/appintro/internal/viewpager/ViewPagerTransformer.kt
```diff
@@ -56,13 +56,16 @@ internal class ViewPagerTransformer(
             }
             TransformType.FADE -> {
                 if (position <= -1.0f || position >= 1.0f) {
+                    page.translationX = page.width.toFloat()
```
cortinico: Again here those should not be here.
Here the fix to your problem: https://help.github.com/en/articles/syncing-a-fork

cortinico:
Please let me know when this is ready for another review, as I saw you're still pushing commit on this branch.

Anutha02:
Ok

Anutha02:
@cortinico Ready for review

Anutha02:
Alright, this was the last commit. This time it is surely ready

PR #670, opened 2019-05-05
Convert AppIntroBase to Kotlin

cortinico:
This is a major refactoring of the AppIntroBase class. I've converted
the class to Kotlin and I've cleaned up a lot of deprecated/unused/dead
code.

I've reduced the surface of the public API to expose only what is really
necessary (the `protected` methods) and make private all the
implementaiton details.

Summary of changes:
* Refactored Callback (removed deprecated callbacks).
* Reduced number of boolean flags (now every flag is a kotlin property
so is easier to understand the side effects).
* Introduced the `onIntroFinished` and `onNextSlide` callback, this
simplifies a bit all the callback code.
* Refactored all the Immersive Mode code to only enable Sticky Immersive
as the non-Sticky made no sense for this library.
* Cleaned up the `checkAndRequestPermissions` to use a more concise functional
approach.


cortinico:
@paolorotolo So this looks like the last file that we have to convert before being 100% Kotlin (at least on the library side :D). We can probably release a `6.0.0-alpha2` or `-beta` just after this as this will have major breaking changes.

@AnuthaDev You probably also want to have a look 👍

AnuthaDev:
@cortinico Oh, I sure want to

AnuthaDev:
@cortinico So far so good. I think it is time we migrate to viewpager2 as the rtl support it brings would help me remove the workarounds applied in  #654. Also, due to the major changes, I think first you should merge this PR. Then we will think over #654 as I have some improvements in mind, so it would be better to submit a new PR

paolorotolo:
Great work @cortinico.
Sure @AnuthaDev we will start to investigate how viewpager2 could help us but I think we may wait till its stable release.

AnuthaDev:
@paolorotolo Well the first roadblock I found was that the Viewpager2 class is final, so we cannot extend it. Hopefully they will change this in the future.

PR #671, opened 2019-05-06
Best Pactise Builder Pattern along with Missing Arabic Word Translations.

Younes-Charfaoui:
Hello there, I've changed two main things:

-  First is Builder Pattern way, I've copied the class SliderPageBuilder to the Main Class SliderPage, that way people know that there is a builder object for this class by Making Just SliderPage.Builder, that how many libraries and android core libraries implement the Builder pattern, I've also refractor its test file, and all tests pass correctly.

- Secondly, I noticed that there were 3 missing words in Arabic translation, so I've added them.


AnuthaDev:
Hey @cortinico What do you think of this?

cortinico:
Hey @Younes-Charfaoui 
Thanks for the contribution :)

The Arabic Translation looks fine, thanks for doing it and we can merge it.

For the `.Builder` class, I'd prefer we keep it as it is (aka can you revert abf4b4b ?).
The reason behind this are:

* `SliderPage` is a Kotlin class. Having a builder there is probably useless as developers can use named parameters on it to specify all the attributes they want.
* The `SliderPageBuilder` is kept for backward compatibility with the Java API (developers that used AppIntro on the past might have used this class). 

So ideally we can probably add some notes on top of `SliderPageBuilder.kt` to explain why the file is there and keep it for backward compatibility.

Younes-Charfaoui:
Hi @cortinico, I did it.

PR #677, opened 2019-05-20
Update dependencies

cortinico:
Just updating all the production dependencies so they don't appear in every PR we receive.


paolorotolo:
Looks good, thanks

PR #676, opened 2019-05-15
Initial Swipe support

AnuthaDev:
Fixes #347 
This PR enables support for swipe even when permissions are requested during the intro.
It is work in progress.
Things currently broken:
1. RTL Support
2. User can sometimes skip slides whose permissions are required.
3. Code optimization.

Also, i think this implementation needs better integration with the existing callbacks, so please guide me on that.
This is a very early PR, so a lot of work needs to be done.

Please send your feedback @cortinico @paolorotolo .

PR #678, opened 2019-05-20
Support requesting permissions without locking the Swipe

cortinico:
Make sure we don't have to `setSwipeLock(true)` if the user is
requesting a permission. This is achieved by requesting the permission
during the `onPageScrolled`. If the permission is denied, AppIntro will
go to the previous slide.

Example: 
May-20-2019 23-28-53

Fix #676
Ping @AnuthaDev

### DIFF
--- a/appintro/src/main/java/com/github/paolorotolo/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/paolorotolo/appintro/AppIntroBase.kt
@@ -3,6 +3,7 @@ package com.github.paolorotolo.appintro
 import android.animation.ArgbEvaluator
 import android.annotation.SuppressLint
 import android.content.Context
+import android.content.pm.PackageManager
 import android.os.Build
 import android.os.Bundle
 import android.os.Vibrator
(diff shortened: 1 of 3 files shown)


AnuthaDev:
@cortinico I will run it ASAP, just a quick suggestion(after looking at the GIF): I think the permissions request should fire when the user tries to leave the permission slide by swiping and not when the slide is opened.

cortinico:
Ideally yes, but that approach is harder to maintain and probably not worth
the effort.
Ideally a "good" intro will:

1) show the user why a permission is requested on slide X
2) Let the user swipe to slide X+1
3) When slide X+1 is selected will request permission
4a) If permission is accepted: user will the slide X+1 with a "thank you
for giving permission"
4b) If permission is denied: user will bump back to slide X with the
explanation on why they should give permission.

Having this approach will also simplify the handling of all the swipe
events when the Permission dialog is displayed (and will simplify the code).

AnuthaDev:
@cortinico Fair Point

AnuthaDev:
@cortinico Another Workaround: We add a button that says"Grant Permissions" on the permission slide and disable swipe right. So, the user selects slide X, then swipes right to go to slide X+1 where swipe right is disabled and a button appears that asks to grant permission.

AnuthaDev:
Something like this: 

https://drive.google.com/file/d/1-7n4X8xUl4LFqFFZo9a2fXO0WD_sfJpG/view?usp=drivesdk

cortinico:
A button "Grant Permission" can always be added by the user of the library
if they provide a custom layout.
The `askForPermissions` method that we provide in AppIntroBase is more like
a utility to simplify Permission handling but the user is free to go custom
if he wants.

AnuthaDev:
@cortinico @paolorotolo I don't think this behaviour should be there. I will submit a new PR that uses my approach in #675 in a much more efficient way.(soon)

paolorotolo:
Sure, feel free to :)

gcantoni:
@cortinico how to implement this? :)

cortinico:
> @cortinico how to implement this? :)

Hey @gcantoni This is currently on master. Will be released with the upcoming version of AppIntro (6.0.0)

PR #681, opened 2019-05-23
Gradle to 5.4.1

cortinico:
Our Gradle is a bit obsolete, let's update it!


AnuthaDev:
@cortinico also update the material drawer dependency to 6.1.2

AnuthaDev:
It supports androidX, so we also don't need jetifier

PR #682, opened 2019-05-23
Remove extra useless gradle.properties

cortinico:


cortinico:
Looks like I was too aggressive 😅 I'm going to fix the top level property file.

PR #214, opened 2016-04-19
[WIP] Feature/extended slide policy readme update, bugfixes

maxee:
Here are a few additions and bugfixes to my last PR:

- Implemented extended slide policy. Slides need to implement `ISlidePolicy` in order to enable this feature. By overriding `isPolicyRespected` a slide can control whether the user should be allowed to go to the last/next slide. By overriding `onUserIllegallyRequestedNextPage` a user can handle illegal slide changing attempts
- Made `ISlideBackgroundHolder` public
- Added slide policy demo to the example app
- Updated default styles for tablets, i.e. increased margin/padding on tablets
- Small bug fixes

PR #689, opened 2019-05-25
AppIntroViewPager: Fix slide policy handling when sliding the view pager

maxee:
Due to lack of overriding `onInterceptTouchEvent` and not overriding `onTouchEvent` correctly, sliding between slides is not getting disabled, even when the associated policy restricts sliding to the next/previous slide.

Somehow my solution provided in 2016/2018 has been exchanged for a wrong one. So I reimplemented the solution again (see #214 )


appintro/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.kt
```diff
@@ -97,45 +96,68 @@ class AppIntroViewPager(context: Context, attrs: AttributeSet) : ViewPager(conte
...
      * Util function to check if the user Illegaly request a swipe.
      * Also checks if the event happened not earlier than every 1000ms
      */
     private fun userIllegallyRequestNextPage(event: MotionEvent): Boolean {
-        if (Math.abs(event.x - currentTouchDownX) >= VALID_SWIPE_THRESHOLD_PX) {
-            if (System.currentTimeMillis() - illegallyRequestedNextPageLastCalled >=
-                ON_ILLEGALLY_REQUESTED_NEXT_PAGE_MAX_INTERVAL
-            ) {
+        if ((event.x - currentTouchDownX).absoluteValue >= VALID_SWIPE_THRESHOLD_PX_X
+                && (event.y - currentTouchDownY).absoluteValue <= VALID_SWIPE_THRESHOLD_PX_Y) {
```
AnuthaDev: Why exactly do we also need Y-Coordinate? Viewpager only supports Horizontal navigation.

maxee: @AnuthaDev It prevents detecting a false positive sliding attempt. Think of a scroll view within a slide. If one wants to scroll the scroll view vertically but also scrolls a bit horizontally (by accident), a sliding attempt would be assumed (if the user scrolled more than 25px horizontally during the intended vertical scroll).

appintro/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.kt
```diff
@@ -97,45 +96,68 @@ class AppIntroViewPager(context: Context, attrs: AttributeSet) : ViewPager(conte
         customScroller?.scrollDurationFactor = factor
     }
 
-    override fun performClick() = super.performClick()
+    override fun onInterceptTouchEvent(event: MotionEvent?): Boolean
+    {
+        if (!handleTouchEvent(event)) {
+            return false
```
cortinico: You need to return `true` here.

appintro/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.kt
```diff
@@ -97,45 +96,68 @@ class AppIntroViewPager(context: Context, attrs: AttributeSet) : ViewPager(conte
...
+        }
+
+        // Calling super will allow the slider to "work" left and right.
+        return super.onInterceptTouchEvent(event)
+    }
 
     override fun onTouchEvent(event: MotionEvent?): Boolean {
+        if (!handleTouchEvent(event)) {
+            return false
```
cortinico: You need to return `true` here.

appintro/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.kt
```diff
@@ -97,45 +96,68 @@ class AppIntroViewPager(context: Context, attrs: AttributeSet) : ViewPager(conte
...
+    private fun handleTouchEvent(event: MotionEvent?) : Boolean {
         // If paging is disabled we should ignore any viewpager touch
         // (also, not display any error message)
         if (!isFullPagingEnabled) {
             return false
         }
 
-        val canRequestNextPage = onNextPageRequestedListener?.onCanRequestNextPage() ?: true
-        when (event?.action) {
-            MotionEvent.ACTION_UP -> performClick()
```
cortinico: Please keep this line for accessibility purposes.

appintro/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.kt
```diff
@@ -97,45 +96,68 @@ class AppIntroViewPager(context: Context, attrs: AttributeSet) : ViewPager(conte
...
     override fun onTouchEvent(event: MotionEvent?): Boolean {
+        if (!handleTouchEvent(event)) {
+            return false
+        }
+
+        // Calling super will allow the slider to "work" left and right.
+        return super.onTouchEvent(event)
+    }
+
+    private fun handleTouchEvent(event: MotionEvent?) : Boolean {
```
lucalves: I would recommend a comment specifying this function.

cortinico:
@maxee Any updates?

appintro/src/main/java/com/github/paolorotolo/appintro/AppIntroViewPager.kt
maxee: Done

maxee:
@cortinico I added another commit, fixing most of the comments, i guess.

cortinico:
> @cortinico I added another commit, fixing most of the comments, i guess.

Great :) How about the other comments? If you don't have time to finish it, I can also pick up this branch from where you left it.

cortinico:
I'm taking over the branch and merging it as this PR was opened long ago and it makes sense to have this fix in master :)

PR #691, opened 2019-05-27
Make Travis run all the Gradle tasks

cortinico:
Currently Travis stops the build as soon as it finds a failure. I'd rather let it run all the tasks so we can have the list of KtLint and Detekt failures.


paolorotolo:
Sounds good

PR #694, opened 2019-06-02
Added Serbian translation

dragstor:
Strings translated to Serbian, within a `values-sr` folder.

### DIFF
--- a/appintro/src/main/res/values-sr/strings.xml
+++ b/appintro/src/main/res/values-sr/strings.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources xmlns:tools="http://schemas.android.com/tools">
+<resources>
     <string name="app_intro_skip_button">PRESKOČI</string>
     <string name="app_intro_next_button">DALJE</string>
     <string name="app_intro_back_button">NAZAD</string>
--- a/appintro/src/main/res/values-sr/strings.xml
+++ b/appintro/src/main/res/values-sr/strings.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
     <string name="app_intro_skip_button">PRESKOČI</string>
-    <string name="app_intro_next_button" tools:ignore="MissingTranslation">DALJE</string>
-    <string name="app_intro_back_button" tools:ignore="MissingTranslation">NAZAD</string>
+    <string name="app_intro_next_button">DALJE</string>
+    <string name="app_intro_back_button">NAZAD</string>
     <string name="app_intro_done_button">KRAJ</string>
-    <string name="app_intro_image_content_description" tools:ignore="MissingTranslation">sličice</string>
+    <string name="app_intro_image_content_description">sličice</string>
 </resources>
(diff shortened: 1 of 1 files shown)


appintro/src/main/res/values-sr/strings.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <string name="app_intro_skip_button">PRESKOČI</string>
+    <string name="app_intro_next_button" tools:ignore="MissingTranslation">DALJE</string>
```
cortinico: Can you remove the `tools:ignore="MissingTranslation"` from this line?

```suggestion
    <string name="app_intro_next_button">DALJE</string>
```

appintro/src/main/res/values-sr/strings.xml
```diff
@@ -0,0 +1,8 @@
...
+    <string name="app_intro_back_button" tools:ignore="MissingTranslation">NAZAD</string>
```
cortinico: Can you remove the `tools:ignore="MissingTranslation"` from this line?

```suggestion
    <string name="app_intro_back_button">NAZAD</string>
```

dragstor:
Note: Strings are translated by using the Latin script, since it's used more than the Cyrillic one (in Apps).
 
How should we handle the Cyrillic script, since both are in use in Serbia?

- Latin script example: 
`Next -> DALJE`
- Cyrillic script example: 
`Next -> ДАЉЕ`

appintro/src/main/res/values-sr/strings.xml
```diff
@@ -0,0 +1,8 @@
...
+    <string name="app_intro_done_button">KRAJ</string>
+    <string name="app_intro_image_content_description" tools:ignore="MissingTranslation">sličice</string>
```
cortinico: Can you remove the `tools:ignore="MissingTranslation"` from this line?

```suggestion
    <string name="app_intro_image_content_description">sličice</string>
```

appintro/src/main/res/values-sr/strings.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:tools="http://schemas.android.com/tools">
```
cortinico: You should be fine also removing the `xmlns:tools` definition here.

dragstor:
@cortinico cleanup done, waiting for Travis and then we're ready 👍 thanks!

cortinico:
> How should we handle the Cyrillic script, since both are in use in Serbia?

IIRC Android doesn't have easy support for IETF BCP 47 (like `sr_Cyrl_RS` “Serbian (Cyrillic, Serbia)”). Ideally we should code something inside the library to handle this. I don't think it's probably worth the effort and I'd rather go with Latin only as you said: _since it's used more than the Cyrillic one (in Apps)._

dragstor:
> > How should we handle the Cyrillic script, since both are in use in Serbia?
> 
> IIRC Android doesn't have easy support for IETF BCP 47 (like `sr_Cyrl_RS` “Serbian (Cyrillic, Serbia)”). Ideally we should code something inside the library to handle this. I don't think it's probably worth the effort and I'd rather go with Latin only as you said: _since it's used more than the Cyrillic one (in Apps)._

I agree. It was just me being curious :) 

Thanks for the merge btw 👍

cortinico:
Thanks to you for the contrib :)

PR #700, opened 2019-06-27
Add methods to allow to change done/skip text by passing String Res ID

fchauveau:


appintro/src/main/java/com/github/paolorotolo/appintro/AppIntro.kt
```diff
@@ -137,8 +158,8 @@ abstract class AppIntro : AppIntroBase() {
      *
      */
     @Deprecated(
-        "use {@link #setProgressButtonEnabled(boolean)} instead.",
-        ReplaceWith("isProgressButtonEnabled = showDone")
+            "use {@link #setProgressButtonEnabled(boolean)} instead.",
```
cortinico: Please amend this change. That's also the reason why the CI it's failing.
Otherwise just run `./gradlew ktlintFormat` and commit the result.

fchauveau: @cortinico fixed! 
(I didnt noticed formatting was broken here sorry)

PR #723, opened 2020-01-03
Add Norwegian translation

perchrh:


cortinico:
I fixed the build here #724 so we should be good to go

PR #686, opened 2019-05-24
New Approach to permissions

AnuthaDev:
Fixes #347

New approach to request permissions.

This PR is more or less complete and ready to merge. I have tested it against various scenarios and it should work fine.

Output: https://drive.google.com/file/d/15K_4EXmh0aq-QYkG9zJ3L7sefnfZYoRJ/view?usp=drivesdk

PR #730, opened 2020-01-23
New Approach to Permission & Cleanup

cortinico:
I'm cherry-picking @AnuthaDev commit plus applying my changes on top of it (so his contribution will be counted in the graph).

This will speedup the review process, as that review was opened in May 2019 and is still around. Doing further review over there is getting complicated.

@AnuthaDev please verify the changes I applied and leave comments if you don't agree on something. Ideally I would love to merge this by the end of this week.
I've done some manually testing and looks like is working fine. Please do further testing.

Closes #686

### DIFF
--- a/appintro/src/main/java/com/github/paolorotolo/appintro/internal/PermissionWrapper.kt
+++ b/appintro/src/main/java/com/github/paolorotolo/appintro/internal/PermissionWrapper.kt
@@ -4,12 +4,12 @@ package com.github.paolorotolo.appintro.internal
  * A data class that represents a set of permissions that should be requested to the user.
  * @property permissions An Array of Permissions from the Android Framework
  * @property position The position in the AppIntro pager when to request those permissions.
- * @property pending Flag set to true if AppIntro already requested this permission and is awaiting response.
+ * @property required Whether the permission being requested needs to be granted to move forward.
  */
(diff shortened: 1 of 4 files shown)


AnuthaDev:
Excellent!! I have tested the changes, the permission stuff is working flawlessly. I approve this PR😁

appintro/src/main/java/com/github/paolorotolo/appintro/internal/PermissionWrapper.kt
```diff
@@ -10,7 +12,7 @@ internal data class PermissionWrapper(
     var permissions: Array<String>,
     var position: Int,
     var required: Boolean = true
-) {
+) : Serializable {
```
AnuthaDev: Just one question, is this Serializable necessary, i couldn't find its use anywhere...

cortinico: It's necessary as we're parcelizing an `HashMap<Int, PermissionWrapper>`. The `HashMap` is actually serializable so you can use `.putSerializable`, but the `PermissionWrapper` is not. It crashes at runtime during the `onResume()` when trying to deserialise such class.

PR #735, opened 2020-01-23
Added a parallax animation setting

AnuthaDev:
This PR adds a parallax animation setting to the library, so that the user can add parallax effects on the views.


cortinico:
Thanks for the contrib @AnuthaDev I've just fixed the KDoc and I'm going to merge this right now.

PR #738, opened 2020-01-23
Fix button state not being retained on configuration change

### LINKED ISSUE #737: [BUG] Appintro Button layout is not retained on configuration change
**AppIntro Version**: latest master branch

**Issue details / Repro steps / Use case background**: Fire up appintro sample app, swipe till you reach last page and the next button turns into done button, now rotate the screen and the done button changes back to next button and does nothing on tap, skip button reappears. Swipe a page back and then forward again, the done button appears again, fully functional, and the skip button disappears.
Screenshot_20200123-144343_AppIntro_Example

cortinico:
As reported by @AnuthaDev, we're not retaining the state of the buttons on configuration change. This PR fixes the bug.

Fixes #737


cortinico:
> Nice. I'm approving the changes anyway since it's not in the scope of this PR, but I would also move Bundle keys into static constants to improve code maintainability in future.

Let's do it! :)

PR #744, opened 2020-01-25
Refactor example package name, update gradle

paolorotolo:
Since 6.0.0 is coming out, this PR refactors example app package name to `com.github.appintro.example` since it was before customized by a previous contributor.

After this lands I'll rewrite the examples in Kotlin to improve readability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,7 +292,7 @@ In order to provide the translation, your file needs to contain the following st
 An updated version of the english version translation is [available here](appintro/src/main/res/values/strings.xml).
 
 ## Example App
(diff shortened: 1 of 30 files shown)


cortinico:
You also have a line to remove from the README:
```
./README.md:See example code [here](https://github.com/AppIntro/AppIntro/tree/master/example) on GitHub. You can also see it live by downloading our example on [Google Play](https://play.google.com/store/apps/details?id=com.amqtech.opensource.appintroexample).
```

paolorotolo:
> You also have a line to remove from the README
Ok I'll remove it for now, but we could publish another example app on one of our personal Play accounts. It's not necessary IMHO to open another Play developer account just for this.

> Do we want to change the package name for the library as well @paolorotolo ?
Hmm since we're making breaking changes here I see no better time to do so. Discussion in #745.

PR #767, opened 2020-04-25
Fix setIndicatorColor crashing onCreate

cortinico:
Currently the setIndicatorColor is crashing on the onCreate
as the indicator is not yet initialized. I'm fixing the initialization
code for the indicator to appen on the top level class onCreate


paolorotolo:
Hmm, I don't know how I feel about checking lateinit initialization... Can we lazily initialize the indicator to avoid this?

cortinico:
> Hmm, I don't know how I feel about checking lateinit initialization... Can we lazily initialize the indicator to avoid this?

Nope we can't use `by lazy{}` as we also offer the possibility to override the indicator (while by lazy is for read only values). I've moved the initialisation in the `onCreate` and made the field nullable.

PR #768, opened 2020-04-25
Refactor Transformers to use a sealed class

cortinico:
I'm cleaning up the API of `AppIntroBase` by removing a lot of methods
used to setup the `PageTransformer`. I'm now exposing only a
`setTransformer` method that accepts an instance of a sealed class.

This allows us to contain a bit the `AppIntroBase` API surface size.


paolorotolo:
Ok for me, thanks

PR #770, opened 2020-04-25
Add missing flags for `setStatusBarColor`

cortinico:
Currently the setStatusBarColor method is not setting the proper
window flags so the color is ignored.
I'm fixing it as well as adding a sample in the sample app.

Also here a screenshot:
Screenshot_1587837157


cortinico:
Let's merge #777 first and I'll rebase on top of that.

paolorotolo:
Done

PR #771, opened 2020-04-25
Update parameter names in SliderPage to be consistent

cortinico:
Currently we have a mixture of names for parameters being long (title,
image) and abbreviation (desc, bg). I'm consolidating all of them to
have the full parameter name across SliderPage, SliderPageBuilder and
AppIntroFragment.

Moreover the sorting of the paramters is not consistent between
SliderPage and AppIntroFragment. I'm fixing it.


paolorotolo:
It seems there's a little conflict with Resources annotation. Could you fix it please?

cortinico:
> It seems there's a little conflict with Resources annotation. Could you fix it please?

Done

PR #773, opened 2020-04-25
Fix bug on swipe with Permission slide

cortinico:
Currently the Swiping on Permission slide is bugged.
When swiping backwards and with a permission slides, it could happen
that the carousel is left "frozen" in between two slides. I'm fixing it.

Moreover I'm also unlocking the slide when the user denied a permission.
It should still be possible for a user to eventually go back in
the carousel even if they rejected a permission.

| Before | After |
| -- | -- |
| untitled | untitled2 |

### DIFF
--- a/appintro/src/main/java/com/github/appintro/internal/AppIntroViewPager.kt
+++ b/appintro/src/main/java/com/github/appintro/internal/AppIntroViewPager.kt
@@ -158,7 +158,6 @@ internal class AppIntroViewPager(context: Context, attrs: AttributeSet) : ViewPa
                 if (isPermissionSlide && isSwipeForward(currentTouchDownX, event.x)) {
                     onNextPageRequestedListener?.onUserRequestedPermissionsDialog()
                 }
-                currentTouchDownX = event.x
             }
         }
         return isFullPagingEnabled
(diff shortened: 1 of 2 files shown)


appintro/src/main/java/com/github/appintro/internal/AppIntroViewPager.kt
```diff
@@ -158,7 +158,6 @@ internal class AppIntroViewPager(context: Context, attrs: AttributeSet) : ViewPa
                 if (isPermissionSlide && isSwipeForward(currentTouchDownX, event.x)) {
                     onNextPageRequestedListener?.onUserRequestedPermissionsDialog()
                 }
-                currentTouchDownX = event.x
```
AnuthaDev, line 161: @cortinico This was added to prevent the next slide from being visible before permissions are granted. In your branch try this:
On slide 2 do a slow half backward swipe and then a rapid full forward swipe, a small portion of the next slide would be visible. In my testing I was actually able to bypass the permission check using this trick. So, .... Be careful while removing this.

AnuthaDev:
You could maybe look at some offsets to decide whether to show the dialog...

cortinico:
> @cortinico This was added to prevent the next slide from being visible before permissions are granted. In your branch try this:
> On slide 2 do a slow half backward swipe and then a rapid full forward swipe, a small portion of the next slide would be visible. In my testing I was actually able to bypass the permission check using this trick. So, .... Be careful while removing this.

I see your point.
I've added a fix to force the recenter of the slide in 67600a6.

I'd also mention that this start to be a edge case scenario, as this will affect only AppIntro users that uses Permissions + have Permissions + mandatory permissions + users that denies such permissions.

If we don't like the fix applied in 67600a6 @paolorotolo we can also skip it and live with it.

The solution that we have in master at the moment is really buggy and it shows this weird scenario whenever you accidentally move one pixel right that is not great imho 🤔

paolorotolo:
Okay, let's merge this for now to fix the buggy behavior of master.
But let's keep an eye on this edge case in future.

PR #774, opened 2020-04-25
Move from Travis to Github Actions

cortinico:
Seems like Travis is a bit unreliable recently.
I'm moving over to Github Actions as it's easier to maintain
and to integrate with our workflow.


cortinico:
For some reason the Run status is not posted in this PR.
Here the link of the run: https://github.com/cortinico/AppIntro/runs/618527813
The integration should work once we merge this PR as in contains the workflow files.

paolorotolo:
First time I actually use GH Actions. I'm curious about the integration with the rest of github.
Let's see!

PR #776, opened 2020-04-27
Renamed AppIntroCustomLayoutSlide -> Fragment

cortinico:
I'm renaming the newly added class `AppIntroCustomLayoutSlide` to `AppIntroCustomLayoutFragment` to be consistent with the other Fragments we offer.


paolorotolo:
Sounds ok for me

PR #802, opened 2020-06-01
Fix goToNextSlide default param for RTL

### LINKED ISSUE #800: wrong default argument when RTL in goToNextSlide() method.
The default argument for isLastSlide on the `goToNextSlide()` method is wrong for RTL layouts.
A correct check for RTL layout's last slide would be checking if currentItem - 1 == 0 (double check me on this one please).

Method's signature: `protected fun goToNextSlide(isLastSlide: Boolean = pager.currentItem + 1 == slidesNumber)`

**AppIntro Version**: 6.0.0

cortinico:
Consolidates the way how we check if we are at the last page (both on RTL or LTR). Currently the default parameter for `goToNextSlide` is wrong for RTL (as reported in #800 and won't work correctly.

Moreover we had the same bug in the keyboard management. I'm fixing it.

Fixes #800

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
@@ -174,7 +174,7 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
     }
 
     /** Moves the AppIntro to the next slide */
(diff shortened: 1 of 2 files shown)


paolorotolo:
Okay, we could plan to publish a bugfix release soon

PR #807, opened 2020-06-02
Extend visibility of isRtl on AppIntroBase to protected

### LINKED ISSUE #806: expose isRtl property
`askForPermissions()` API requires a slide number. 
The slide numbers differs when the layout is in RTL or not.
That means I have to check if the layout is RTL, and if it is, the slide number will differ when it isn't.
I'm aware of that the logic is fairly simple, but it would be nicer if it would just have `protected` visibility to prevent multiple implementations across the app :)

**AppIntro Version**: 6.0.0

cortinico:
Developers might need to know if we're on RTL to provide the correct slide number for the `askForPermissions` method.

Fixes #806


cortinico:
Closing as it's not needed. See #806

Issue #801, opened 2020-06-01
Integration AppIntro into Androidx navigation component


oliexdev:
**AppIntro Version**: 6.0.0

**Device/Android Version**: Pixel 2

**Issue details / Repro steps / Use case background**: 

For the new [Androidx navigation framework](https://developer.android.com/guide/navigation/navigation-getting-started) the activity needs a navHost fragment. Is it possible to add that into AppIntro?

Or could the AppIntro class also inherited from a Fragment class?

The problem that I am facing is that I want to navigate to another Fragment (outside of the AppIntro acitivty) in my main nav graph with safe arguments but I can't do it without having a second graph in AppIntro or disabling the second AppIntro activity.

Does any have an idea how to solve this issue?

cortinico:
Thanks for raising this issue @oliexdev 

Currently there is no direct support for the Navigation framework. We definitely need to look into this and provide better guidance on how to integrate an AppIntro Activity inside a navigation graph.

> Or could the AppIntro class also inherited from a Fragment class?

Not possible at the moment and also really unlikely in the foreseeable future.

> The problem that I am facing is that I want to navigate to another Fragment (outside of the AppIntro acitivty) in my main nav graph with safe arguments but I can't do it without having a second graph in AppIntro or disabling the second AppIntro activity.

Can ask you to be more clear here?

> Does any have an idea how to solve this issue?

Ideally treat your whole AppIntro Activity as a separate entity. Once you started it, you let the user finish the flow. Once the AppIntro is over, you can navigate to the entry point of your nav graph.

oliexdev:
Thank for your quick reply. :trophy: 

> Can ask you to be more clear here?

I have a UserSetting fragment class which is used in my app. During the AppIntro slide I added a custom button which should open the UserSetting fragment. So I can add the user during the app intro slide, see [my app](https://f-droid.org/repository/browse/?fdid=com.health.openscale) for an example.

> Ideally treat your whole AppIntro Activity as a separate entity. Once you started it, you let the user finish the flow. Once the AppIntro is over, you can navigate to the entry point of your nav graph.

That will only work if I don't interact with my other fragments but the goal is to interact with other fragments outside of the AppIntro activity using the new Androidx navigation component.

cortinico:
> but the goal is to interact with other fragments outside of the AppIntro activity using the new Androidx navigation component

Yeah that's exactly the point that is not supported at the moment.

WolandPL:
[...]

PR #810, opened 2020-06-03
Basic Androidx navigation component support

oliexdev:
regarding to issue #801 I implemented a basic support for the new Androidx navigation component, so that an app can set a navigation graph and navigate to another fragment (outside of the AppIntro activity)

This PR is **not** merge ready!

This should show only the basic principle for an navigation sample.

What still have to be done is the following:

- Separate AppIntro class into a AppNav class and a AppInto fragment. The AppNav class should only contain the general layout (like toolbar and so on). The AppIntro fragment should contain the ViewPager. This fragment could be the default fragment which is shown into the NavFragment.
(Currently in this PR for a simplification I divide the screen one half for the NavFragment and the lower part for the ViewPager)

- Convert the current Java classes into Kotlin (Sorry I never used Kotlin before, so I programmed in Java)

- Currently, I didn't found a solution that the SafeArgs method works but you could pass Arguments with the Bundle method to another Fragment

If you have any question, please feel free to ask me. I will try to answer them.

### DIFF
--- a/appintro/build.gradle
+++ b/appintro/build.gradle
@@ -3,6 +3,7 @@ apply plugin: 'kotlin-android'
 apply plugin: 'com.github.dcendents.android-maven'
 apply plugin: 'io.gitlab.arturbosch.detekt'
 apply plugin: 'org.jlleitschuh.gradle.ktlint'
+apply plugin: "androidx.navigation.safeargs"
 
 group = 'com.github.AppIntro'
 
@@ -29,6 +30,8 @@ dependencies {
     testImplementation 'junit:junit:4.13'
     testImplementation 'org.mockito:mockito-core:2.28.2'
     implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
+    implementation 'androidx.navigation:navigation-fragment:2.3.0-beta01'
+    implementation 'androidx.navigation:navigation-ui:2.3.0-beta01'
 }
 
 repositories {
--- a/appintro/src/main/res/layout/appintro_intro_layout.xml
+++ b/appintro/src/main/res/layout/appintro_intro_layout.xml
@@ -9,6 +9,25 @@
     android:fitsSystemWindows="false"
     android:theme="@style/AppIntroStyle">
 
+    <fragment
+        android:id="@+id/nav_appintro_fragment"
+        android:name="androidx.navigation.fragment.NavHostFragment"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        app:defaultNavHost="false"
+        app:layout_constraintBottom_toTopOf="@+id/guideline"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:navGraph="@navigation/appintro_navigation" />
+
+    <androidx.constraintlayout.widget.Guideline
+        android:id="@+id/guideline"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:orientation="horizontal"
+        app:layout_constraintGuide_percent="0.5" />
+
     <com.github.appintro.internal.AppIntroViewPager
         android:id="@+id/view_pager"
         android:layout_width="0dp"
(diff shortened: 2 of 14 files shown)


cortinico:
Thanks for the PR @oliexdev, really appreciate.
There are several points that should be discussed:

> Separate AppIntro class into a AppNav class and a AppInto fragment

Adding another layer to the view hierarchy seems like not the best approach. We currently already have an Activity and several fragments in the ViewPager. Adding another intermediate Fragment just to support the Navigation library seems not really worth to me.

Ideally the whole Navigation integration should be isolated inside an `appintro-navigation` module and it should be optional. I don't even know if this is achievable at the current status.

> so that an app can set a navigation graph and navigate to another fragment (outside of the AppIntro activity)

If I understand correctly, you just want to navigate outside of a specific slide of the AppIntro. Can't you just `startActivity` there with the activity you want to start.

I feel the real benefit of providing a potential navigation integration is that developers can deep-link inside a specific page of the AppIntro (so the opposite of your use case). That would require that navigation destination are generated properly for every slide, etc.

I don't know how much effort would be to develop such solution, but it feels to me that currently it's probably not worth the time. 

Side note: the PR is also not building so I wasn't really able to give it a try.

oliexdev:
> Adding another layer to the view hierarchy seems like not the best approach. We currently already have an Activity and several fragments in the ViewPager. Adding another intermediate Fragment just to support the Navigation library seems not really worth to me.
> 
> Ideally the whole Navigation integration should be isolated inside an appintro-navigation module and it should be optional. I don't even know if this is achievable at the current status.

Sure that's a design decision which is up to the maintainer. I never worked before with gradle modules, so I can't really help in this goal.

> If I understand correctly, you just want to navigate outside of a specific slide of the AppIntro. Can't you just startActivity from there with the activity you want to start.
> 
> I feel the real benefit of providing a potential navigation integration is that developers can deep-link inside a specific page of the AppIntro (so the opposite of your use case). That would require that navigation destination are generated properly for every slide, etc.
> 
> I don't know how much effort would be to develop such solution, but it feels to me that currently it's probably not worth the time.

Not really as in issue #801 described it would break the navigation component in the main app because it is recommend to have only one activity and the other UI are fragments. So you can jump out with startAcitivty but only if you wrap the other fragments within another activity in the main app as I did in my current workaround implementation at commit https://github.com/oliexdev/openScale/commit/83e55e05bc9d93f4ba4df7a64b3a44b8b70710b1
But with this you break the navigation component features. Normally you can navigate directly to a fragment with optional safeArgs and other features which comes with Androidx. If you want to jump into an AppIntro activity (the other way around) with the navigation component that is already possible but idk how to select a specific slide in the AppIntro.

> Side note: the PR is also not building so I wasn't really able to give it a try.

The build error were in your original code, see https://github.com/AppIntro/AppIntro/pull/810/commits/d4b8b5b1ed0a436a1ca2c6d173773371f7d8f56a but I fixed it. :wink:

cortinico:
> Not really as in issue #801 described it would break the navigation component in the main app because it is recommend to have only one activity and the other UI are fragments.

Can you elaborate more here? I don't see how not supporting Navigation in AppIntro will break your app. I understand that is recommended to follow a single activity approach but that's not required, especially if you rely on third party libraries such as AppIntro

oliexdev:
I am sorry, I think I can't elaborate it more as I wrote:

> So you can jump out with startAcitivty but only if you wrap the other fragments within another activity in the main app as I did in my current workaround implementation at commit oliexdev/openScale@83e55e0
> But with this you break the navigation component features. Normally you can navigate directly to a fragment with optional safeArgs and other features which comes with Androidx. If you want to jump into an AppIntro activity (the other way around) with the navigation component that is already possible but idk how to select a specific slide in the AppIntro.

What in detail you don't understand? Maybe you see it better if you try by yourself to navigate outside AppInto to an already existing fragment outside AppInto which use Androidx navigation component? I think the use cases that you have to jump out of AppIntro activity for e.g. to create a user is not low.

cortinico:
> I think the use cases that you have to jump out of AppIntro activity for e.g. to create a user is not low.

I see your point. Unfortunately AppIntro doesn't support this and we're probably not going to support it.

Our API is already quite extensive, and supporting navigating outside of a started AppIntro is currently not supported.

Moreover, you're the first user requesting this feature so far.

oliexdev:
no problem I can fully understand that this feature is too troublesome to implement and maintain if I am the only one with this use case. I can live with my current workaround solution.

cortinico:
> no problem I can fully understand that this feature is too troublesome to implement and maintain if I am the only one with this use case.

Thanks anyway for the contribution. 
We will keep issue #801 opened so if someone else has a similar setup as yours and want to try a different approach, we can re-discuss it 👍

PR #816, opened 2020-06-08
Cleanup the Sample App

cortinico:
Time to cleanup a bit our sample app. In this PR I've:

- Removed the last .java files we had in the repo
- Removed the dependency on Material Drawer and removed the ViewPager inside the sample app.
- Removed a lot of old/unused resources
- Removed the stored APK as they're probably really old + we should store them in the artifact instead
- Renamed the app id to `com.github.appintro.example`
- Setup a layout to use a simple RecyclerView with a list of items and buttons.

I think that's a good compromise where we can start building around if needed (like add sections or so).

This reduces the amount of code in the sample app, making it easier for users to navigate through and find the actual examples to copy-n-paste.

The output should look like this:

<p align="center">
    appintro sample app
</p>


paolorotolo:
Great work! I'll give it a go and let you know asap :)

Issue #803, opened 2020-06-02
Back swipe not permitted when slide policy is not respected


idish:
For a given slide with a slide policy that is not respected, swiping back won't work. 
I would expect that only swiping forward won't be permitted.
Problem relies within the following method:
```kotlin
    private fun userIllegallyRequestNextPage(event: MotionEvent): Boolean {
        if (isASwipeGesture(event, currentTouchDownX, currentTouchDownY) &&
            System.currentTimeMillis() - illegallyRequestedNextPageLastCalled >=
            ON_ILLEGALLY_REQUESTED_NEXT_PAGE_MAX_INTERVAL
        ) {
            illegallyRequestedNextPageLastCalled = System.currentTimeMillis()
            return true
        }

        return false
    }
``` 
IMO that method should make use of the existing `isSwipeForward()` util method as used for permission slides as well.

**AppIntro Version**: 6.0.0

cortinico:
IIRC this was kept for compatibility with the behavior that the `ISlidePolicy` interface was having in AppIntro 4.x

It probably makes sense to update it to allow back-swiping.

idish:
@cortinico Yep, I'm aware that it was the same behaviour since then, But I see no reason to disable back swiping.

idish:
IMHO that should be labeled more as a bug, than a feature request.

PR #817, opened 2020-06-08
enabling backward sliding when slide policy not met

idish:
Fix issue https://github.com/AppIntro/AppIntro/issues/803

Now even though that the slide policy is not met, backward sliding will be permitted.


cortinico:
I've added a cleanup for the `isASwipeGesture` as that check is not needed anymore (the `isSwipeForward` check is stricter).

Fixes #803

cortinico:
Thanks for your contribution 👍

PR #825, opened 2020-06-13
Added ability to color back arrow and done button

beefsausage:
Added `setBackArrowColor` and `setDoneButtonColor` to change the color of the back arrow and the background color of the done button.

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntro.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntro.kt
@@ -39,8 +39,8 @@ abstract class AppIntro : AppIntroBase() {
      * @param color your color
      */
     fun setBackArrowColor(@ColorInt color: Int) {
-        val nextButton = findViewById<ImageButton>(R.id.back)
-        nextButton.setColorFilter(color)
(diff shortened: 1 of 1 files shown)


appintro/src/main/java/com/github/appintro/AppIntro.kt
```diff
@@ -33,6 +33,16 @@ abstract class AppIntro : AppIntroBase() {
         nextButton.setColorFilter(color)
     }
 
+    /**
+     * Override back button arrow color
+     *
+     * @param color your color
+     */
+    fun setBackArrowColor(@ColorInt color: Int) {
+        val nextButton = findViewById<ImageButton>(R.id.back)
```
cortinico, line 42:
```suggestion
        val backButton = findViewById<ImageButton>(R.id.back)
```

cortinico:
Thanks for the contribution @beefsausage 👍  We need one small change and we're good to go

beefsausage:
Yeah adding the done button color option does create an inconsistency, I removed it and just fixed the typo. 
Thanks for the review!

Issue #826, opened 2020-06-17
Readme improvements


AnuthaDev:
Hello again! Just some suggestions:

1. Put the appintro logo above the title, replace its white background by transparent (dark mode)(It might be quite difficult considering how I created the illustration... Sorry!😅😅), reduce its size.

2. IMO the slide transformer animations should be in a 3x2 grid to make better use of the screen width and require less scrolling

Basically, the same kind of layout as https://github.com/heinrichreimer/material-intro 😁

AnuthaDev:
Also, could you please provide an apk file of the sample app, like we had before? That'd be great

cortinico:
Closing as everything has been addressed by #827 and #828. The logo changes you suggested was already discussed in #732 and we ended up for the logo/gif side by side.

PR #827, opened 2020-06-17
Put slide transformers side by side in README

cortinico:
As suggested in #826, let's cleanup a bit the transformer table as it's too big.


paolorotolo:
Much better, thanks

Issue #841, opened 2020-08-11
isPolicyRespected() also prevents user from swiping back to the previous slide *JAVA*


LassassinX:
**AppIntro Version**: 6.0.0

**Device/Android Version**: SAMSUNG SM-A20F

**Issue details / Repro steps / Use case background**: Hello, i wish to prevent my user from accessing the next slide unless they press a button. However, using the isPolicyRespected() also prevents the user from swiping to the previous slide as well. When in wizard mode, I can use the back button to go back to the slide, but I cannot swipe and go back to the slide. Anyway to fix this ? I am using java

EDIT: I have tried to use the setNextPageSlideLock(true), but it doesn't work. Putting setSwipeLock(true) in the onSlideChanged() method works, but that is not what I'm looking for, as it disables both forward and reverse swiping. I want my users to be able to swipe back, but not swipe forwards. The updated code is below 

**Your Code**:  
//MainActivity
        
       @Override
        protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        welcomeSlide = WelcomeSlide.newInstance(ColorEx.WHITE);
        loginSlide = LoginSlide.newInstance(ColorEx.WHITE);
        teacherOrStudentSlide = TeacherOrStudentSlide.newInstance(ColorEx.TUTOR_BEAR_DARK_BLUE);
        createMyAccountSlide = CreateMyAccountSlide.newInstance(ColorEx.TUTOR_BEAR_DARK_BLUE);

        addSlide(welcomeSlide);
        addSlide(loginSlide);
        addSlide(teacherOrStudentSlide);
        addSlide(createMyAccountSlide);

        setColorTransitionsEnabled(true);

        setWizardMode(true);

        setSkipButtonEnabled(false);

        setDoneText("");

        changeIndicatorColor("Dark Blue");

    }

    @Override
    protected void onSlideChanged(Fragment oldFragment, Fragment newFragment) {
        super.onSlideChanged(oldFragment, newFragment);
        if (newFragment == loginSlide) {
            changeIndicatorColor("Dark Blue");
        }

        if (newFragment == teacherOrStudentSlide) {
            setNextPageSwipeLock(true);
            changeIndicatorColor("White");
        }

    }

//TeacherOrStudent class

      private static final String ARG_COLOR_INT = "COLOR_INT";
    private int originalColor = 0;
    private ConstraintLayout background = null;

    @Override
[...]

PR #847, opened 2020-08-19
Add a sample with SlidePolicy to the example app

cortinico:
I'm adding a sample to the sample app with `SlidePolicy`. 
This helps us to clarify the #841 issue and provides a simple copy-n-paste solution for a `SlidePolicy` that we can point to our users.

Here a video of the sample:
untitled

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,6 +436,8 @@ class MyFragment : Fragment(), SlidePolicy {
 }
 ```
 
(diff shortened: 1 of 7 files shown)


paolorotolo (approved):
Looks good! If you're ok with it, I would just add some words in README in the Slide Policy section like "A full working sample is available [here](link to sources).

cortinico:
> Looks good!
> If you're ok with it, I would just add some words in README in the Slide Policy section like "A full working sample is available [here](link to sources)".

Yup added 👍

PR #852, opened 2020-08-25
Function for setting text size of Done and Skip buttons

bmalex88:
I found myself in need of this one. The other alternative was overriding the entire `AppIntroButtonStyleCompat` style

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntro.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntro.kt
@@ -1,13 +1,12 @@
 package com.github.appintro
 
 import android.graphics.drawable.Drawable
+import android.util.TypedValue
 import android.view.View
 import android.widget.ImageButton
 import android.widget.ImageView
 import android.widget.TextView
-import androidx.annotation.ColorInt
-import androidx.annotation.FontRes
-import androidx.annotation.StringRes
+import androidx.annotation.*
 import com.github.appintro.internal.TypefaceContainer
 
 abstract class AppIntro : AppIntroBase() {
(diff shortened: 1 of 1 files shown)


cortinico:
Thanks for your PR @bmalex88 
Please make sure you run detekt and ktlint as the CI is red

appintro/src/main/java/com/github/appintro/AppIntro.kt
```diff
@@ -163,6 +162,18 @@ abstract class AppIntro : AppIntroBase() {
         nextButton.setImageDrawable(imageNextButton)
     }
 
+    /**
+     * Override done and skip buttons text size
+     *
+     * @param sizeText your text size from resource (in sp)
+     */
+    fun setTextSizeButtons(@DimenRes sizeText: Int) {
```
cortinico, line 170: This should be split in two methods to be consistent with the rest of the API.

Moreover please change this to accept a TextApperance:

`fun setDoneTextAppearance(@StyleRes textAppearance) {`
`fun setSkipTextAppearance(@StyleRes textAppearance) {`

cortinico (changes requested):

appintro/src/main/java/com/github/appintro/AppIntro.kt
bmalex88: I thought about changing the appearance altogether at first so I will do that. Separating them makes sense from that point of view, although I don't believe people are going to change their styles separately

cortinico: > although I don't believe people are going to change their styles separately

Agree, but at the same time we offer both `setDoneTextTypeface` and `setSkipTextTypeface` so we should keep the API consistent on this front.

PR #856, opened 2020-08-27
Fix blending of Indicator Colors

### LINKED ISSUE #855: setIndicatorColor does not change indicator color unless I override appintro_default_unselected_color
There's some strange behavior happening when using `setIndicatorColor`. It does not appear to change the dot indicators' colors. Only if I override `appintro_default_unselected_color` in my _colors.xml_ file it works. But if I override the color yet not use `setIndicatorColor`, it does not work either.

**AppIntro Version**: 6.0.0
**Device/Android Version**: Galaxy S8

Use `setIndicatorColor` without overriding `appintro_default_unselected_color` or override `appintro_default_unselected_color` without using `setIndicatorColor` and the color of the indicator dots does not change.

**Your Code**: 

    setIndicatorColor(
        ContextCompat.getColor(this, R.color.selectedIndicatorColor),
        ContextCompat.getColor(this, R.color.unselectedIndicatorColor)
    )

In _colors.xml_ I have to add

    <color name="appintro_default_unselected_color">#79cee4</color>

cortinico:
Currently, we blend the user provided colors for the Dot indicator with `appintro_default_unselected_color`. That color has a transparency leading to sometimes unexpected behavior.

I'm cleaning up the way how we blend colors for indicators. Now the indicator has a white background and is tinted using:
- `appintro_default_unselected_color` for unselected indicator, if not specified
- `appintro_default_selected_color` for selected indicator, if not specified
- The user provided colors will override those colors and no blending will happen.

I've also cleaned up the code a bit as we were using some deprecated functions + we were re-inflating the indicator drawable for every call to `selectPosition`. Now is way more streamlined.

Fixes #855


paolorotolo (approved):
Looks good, thanks!

PR #857, opened 2020-08-27
Function for setting text appearance of Done and Skip buttons

bmalex88:
This time the two functions allow changing the entire appearance of the Done and Skip buttons

### DIFF
--- a/appintro/src/main/res/layout/appintro_intro_layout.xml
+++ b/appintro/src/main/res/layout/appintro_intro_layout.xml
@@ -49,7 +49,7 @@
 
     <ImageButton
         android:id="@+id/back"
-        style="@style/AppIntroImageButtonStyleCompat"
+        style="@style/AppIntroButtonStyleCompat"
         android:contentDescription="@string/app_intro_back_button"
         android:rotation="180"
         android:visibility="invisible"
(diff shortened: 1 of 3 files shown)


bmalex88:
I also found an issue after increasing the text size for **DONE** and **SKIP**. The arrows are no longer vertically centered if I do so.

bottom-displaced

Inexplicably somehow, if the style for the **DONE** and **SKIP** is separated from the the style of the **NEXT** and **BACK** buttons, everything is fixed.

bottom-centered

This also allows for separate customization of the _text buttons_ from the _image views buttons_.

cortinico:
> I also found an issue after increasing the text size for **DONE** and **SKIP**. The arrows are no longer vertically centered if I do so.

Thank you very much for your findings @bmalex88 
Can I ask you to separate this in another PR (as one is technically a bugfix and the other is a new feature).

appintro/src/main/res/layout/appintro_intro_layout.xml
```diff
@@ -33,6 +33,7 @@
         android:id="@+id/bottom"
         android:layout_width="match_parent"
         android:layout_height="@dimen/appintro_bottombar_height"
+        android:padding="0dp"
```
cortinico, line 36: Why is this needed?

cortinico (approved):
LGTM 👌  Thank you very much for the contrib

PR #858, opened 2020-08-28
Update Readme with PrezziBenzina

vernazza:
Add PrezziBenzina to the list of apps using AppIntro


cortinico (approved):

cortinico:
Thank you very much 🙏

PR #861, opened 2020-08-29
Update support channel from Gitter to Slack

cortinico:
As discussed internally, we're updating our primary support channel from Gitter to Kotlin lang Slack.
We believe that this will move us closer to other Kotlin/Android projects and potentially help us to get more involvement.


paolorotolo (approved):
Sure!

PR #863, opened 2020-09-01
Bottom button styles separated from image buttons

bmalex88:
In a separate PR, as requested


cortinico:
Thanks for making another PR as requested.

Regarding this: 
https://github.com/AppIntro/AppIntro/pull/857#issuecomment-682190797

I've actually done more testing and I wasn't able to reproduce the behavior you described. I'm also unsure this specific PR is needed. Can you please investigate a bit further and provided a reproducible example?

cortinico:
Closing as this API is not longer needed

PR #869, opened 2020-09-14
change skip and next arrow button color

Hoossayn:


example/src/main/java/com/github/appintro/example/ui/default/DefaultIntro2.kt
```diff
@@ -53,6 +54,8 @@ class DefaultIntro2 : AppIntro2() {
                 backgroundDrawable = R.drawable.back_slide5
         ))
 
+        setNextArrowColor(Color.RED)
+        setSkipArrowColor(Color.BLUE)
```
cortinico, line 58: Can you move those two to either `CustomBackgroundIntro.kt`?
The `DefaultIntro2` is used for screenshots in the README and we would rather keep it 
less customized.

cortinico (approved):
Thanks for your contribution @Hoossayn 
I left just a minor comment.
You also need to run ktlint (with `./gradlew ktlintFormat`) to make sure that the build is green ✅

ch22843:
Hi. There is no info in readme guide. please add it ) thank you

PR #871, opened 2020-09-26
Added Java Example for App Intro

### LINKED ISSUE #850: Add Java example to the sample app
We should consider adding one intro example in Java inside the example App.
We keep on receiving several support questions that are Java related. Ideally having one sample in Java could provide an easier example that users can use.

chetanmuliya:
Fixes #850 

Added JavaExample app in an example folder


cortinico:
> Added JavaExample app in an example folder

Thanks for the contribution @chetanmuliya 
I would suggest we try to follow a different approach. How about we instead have a `java-appintro` module inside the `example` app and we just have a Java example there? 
This PR is adding 1K+ Loc and it's going to be harder to maintain in the long run.

chetanmuliya:
@cortinico  what if I add java file below kotlin file ex: in default package one is DefaultIntro.kt and another Defaultintro.java ?

cortinico:
> in default package one is DefaultIntro.kt and another Defaultintro.java in same example module ?

That would also work. Adding a `JavaIntro.java` would be a good starting point 👍

chetanmuliya:
Yes, Thank you

chetanmuliya:
I added one java file in default folder and I renamed default package to defaullt because it was creating a complication to add java file

cortinico:
> @cortinico I added one java file in default folder and I renamed default package to defaullt because it was creating a complication to add java file

I suggest you revert this change. `defaullt` looks like a typo in the code. The package for java can just be called `.java` instead of `.default`.

chetanmuliya:
@cortinico added package .java and added file javaintro.java review it and inform me so that I can add further  java files

example/src/main/java/com/github/appintro/example/ui/java/JavaIntro.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class JavaIntro extends AppIntro {
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+
+        addSlide(AppIntroFragment.newInstance("Welcome!",
+                "This is a demo of the AppIntro library, with a custom background on each slide!",
```
cortinico, line 21: Maybe this can be customized to mention that this is an example in Java?

example/src/main/java/com/github/appintro/example/ui/java/JavaIntro.java
```diff
@@ -0,0 +1,54 @@
...
+                "The library offers a lot of customization, while keeping it simple for those that like simple.",
+                R.drawable.ic_slide3
+        ));
+
+        addSlide(AppIntroFragment.newInstance(
+                "Explore",
+                "Feel free to explore the rest of the library demo!",
+                R.drawable.ic_slide4
+        ));
+    }
```
cortinico, line 41: Can you add a `setTransformer(...)` here as it's the most asked question for Java?

example/src/main/java/com/github/appintro/example/ui/java/JavaIntro.java
```diff
@@ -0,0 +1,54 @@
...
+
+import androidx.annotation.Nullable;
+import androidx.fragment.app.Fragment;
+
+import com.github.appintro.AppIntro;
+import com.github.appintro.AppIntroFragment;
+import com.github.appintro.example.R;
+
+
+public class JavaIntro extends AppIntro {
```
cortinico, line 13: You need to add this class inside: https://github.com/AppIntro/AppIntro/blob/master/example/src/main/java/com/github/appintro/example/ui/IntroEntry.kt

Otherwise it won't show up in the menu in the sample app.

cortinico (changes requested):

example/src/main/java/com/github/appintro/example/ui/java/JavaIntro.java
chetanmuliya: done

example/src/main/java/com/github/appintro/example/ui/java/JavaIntro.java
chetanmuliya: added in java @cortinico

chetanmuliya:
@cortinico I did all the changes as you mentioned earlier

example/src/main/java/com/github/appintro/example/ui/IntroEntry.kt
```diff
@@ -18,6 +19,7 @@ data class IntroEntry(
 
 val defaultEntries = listOf(
         IntroEntry(R.string.default_intro_title, R.string.default_intro, DefaultIntro::class.java),
+        IntroEntry(R.string.default_java_intro_title, R.string.default_java_intro, JavaIntro::class.java),
```
cortinico, line 22: Can you move it to the end?

example/src/main/java/com/github/appintro/example/ui/default/DefaultIntro.kt
```diff
@@ -4,6 +4,7 @@ import android.os.Bundle
 import androidx.fragment.app.Fragment
 import com.github.appintro.AppIntro
 import com.github.appintro.AppIntroFragment
+import com.github.appintro.AppIntroPageTransformerType
```
cortinico, line 7: This can be removed

example/src/main/java/com/github/appintro/example/ui/default/DefaultIntro.kt
```diff
@@ -33,6 +34,7 @@ class DefaultIntro : AppIntro() {
                 "Feel free to explore the rest of the library demo!",
                 imageDrawable = R.drawable.ic_slide4
         ))
+
```
cortinico, line 37: This can be removed

example/src/main/java/com/github/appintro/example/ui/java/JavaIntro.java
```diff
@@ -0,0 +1,67 @@
...
+                R.drawable.ic_slide3
+        ));
+
+        addSlide(AppIntroFragment.newInstance(
+                "Explore",
+                "Feel free to explore the rest of the library demo!",
+                R.drawable.ic_slide4
+        ));
+
+        /**You can add any one of the Transition*/
```
cortinico, line 43: I think that having only one is sufficient 👍

cortinico (changes requested):
Thanks for the follow up 👍 There are still a couple of things to fix but we're mostly good to go

chetanmuliya:
Thank you so much @cortinico for giving this opportunity  to contribute to this library

cortinico (approved):
Great 👍 Thanks for the follow ups

Issue #504, opened 2017-11-14
Do not show position dot if there is only one slide


andreynovikov:
I'm constructing slide queue dynamically based on various conditions. It may be a situation when there is only one slide available to user. In that case a dot (position indicator) is superfluous. I propose to make it not visible in that case.

SunshineNumber1:
have you resovel this question? i have the same situation for fix;

SunshineNumber1:
try this:
setIndicatorColor(Color.parseColor("#00000000"),Color.parseColor("#00000000"));

cortinico:
Closed by #876

PR #876, opened 2020-10-01
Single slide indicator condition issue504

Debanshu777:
#504 > Do not show position dot if there is only one slide

I resolved it with the issue by adding two conditions in
..\AppIntro\appintro\src\main\java\com\github\appintro\indicator\DotIndicatorController.kt

### DIFF
--- a/appintro/src/main/java/com/github/appintro/indicator/DotIndicatorController.kt
+++ b/appintro/src/main/java/com/github/appintro/indicator/DotIndicatorController.kt
@@ -44,7 +44,6 @@ class DotIndicatorController(context: Context) : IndicatorController, LinearLayo
 
     override fun initialize(slideCount: Int) {
         this.slideCount = slideCount
-
         for (i in 0 until slideCount) {
             val dot = ImageView(this.context)
             dot.setImageDrawable(
@@ -70,12 +69,7 @@ class DotIndicatorController(context: Context) : IndicatorController, LinearLayo
             } else {
                 unselectedIndicatorColor
             }
-            (getChildAt(i) as ImageView).let {
-                DrawableCompat.setTint(it.drawable, tint)
-            }
-            if (slideCount == 1) {
-                (getChildAt(i) as ImageView).visibility = View.INVISIBLE
-            }
+            (getChildAt(i) as ImageView).let { DrawableCompat.setTint(it.drawable, tint) }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


appintro/src/main/res/values-bn-rIN/strings.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="app_intro_skip_button">এড়িয়ে যান</string>
```
cortinico, line 3: This file should be outside of your diff

Debanshu777:
I have added the changes as suggested @cortinico 
> This file should be outside of your diff
Reverted the last two commits 8f695df & 1a07479 and adding a new modified commit b327651

appintro/src/main/java/com/github/appintro/indicator/DotIndicatorController.kt
```diff
@@ -46,6 +46,7 @@ class DotIndicatorController(context: Context) : IndicatorController, LinearLayo
         this.slideCount = slideCount
 
         for (i in 0 until slideCount) {
+            if (slideCount == 1) { break }
```
cortinico, line 49: I would suggest you instead change this to (you need to move it down):

```suggestion
            if (slideCount == 1) { 
                dot.visibility = View.INVISIBLE
            }
```

This will make sure that the dot is still added to the hierarchy and it's hidden.
You can then amend the change to lines 71-73

cortinico (changes requested):

appintro/src/main/java/com/github/appintro/indicator/DotIndicatorController.kt
Debanshu777: Okay I got the point, I will make the required changes.

appintro/src/main/java/com/github/appintro/indicator/DotIndicatorController.kt
```diff
@@ -67,7 +70,12 @@ class DotIndicatorController(context: Context) : IndicatorController, LinearLayo
             } else {
                 unselectedIndicatorColor
             }
-            (getChildAt(i) as ImageView).let { DrawableCompat.setTint(it.drawable, tint) }
+            (getChildAt(i) as ImageView).let {
+                DrawableCompat.setTint(it.drawable, tint)
+            }
+            if (slideCount == 1) {
+                (getChildAt(i) as ImageView).visibility = View.INVISIBLE
+            }
```
cortinico, line 78:
```suggestion
            (getChildAt(i) as ImageView).let { DrawableCompat.setTint(it.drawable, tint) }
```

Those changes are unnecessary and can probably be removed.

cortinico (approved):
Thanks for the contribution @Debanshu777 🙏

cortinico:
Fixes #504

PR #877, opened 2020-10-04
Tamil translation added

manimaran96:


appintro/src/main/res/values-ta/strings.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <string name="app_intro_skip_button">தவிர்</string>
+    <string name="app_intro_next_button" tools:ignore="MissingTranslation">அடுத்து</string>
```
cortinico, line 4:
```suggestion
    <string name="app_intro_next_button">அடுத்து</string>
```

appintro/src/main/res/values-ta/strings.xml
```diff
@@ -0,0 +1,8 @@
...
+    <string name="app_intro_back_button" tools:ignore="MissingTranslation">பின்னால்</string>
```
cortinico, line 5:
```suggestion
    <string name="app_intro_back_button">பின்னால்</string>
```

appintro/src/main/res/values-ta/strings.xml
```diff
@@ -0,0 +1,8 @@
...
+    <string name="app_intro_done_button">முடிந்தது</string>
+    <string name="app_intro_image_content_description" tools:ignore="MissingTranslation">வரைகலை</string>
```
cortinico, line 7:
```suggestion
    <string name="app_intro_image_content_description">வரைகலை</string>
```

cortinico (changes requested):
Thanks for the contribution 🙏  You just need to remove the `tools:ignore="MissingTranslation"` and we can merge it.

appintro/src/main/res/values-ta/strings.xml
manimaran96: Thanks @cortinico. 
Now updated.

cortinico (approved):

PR #883, opened 2020-10-21
Changing Visibility of Progress Bar when count of silde is one

Debanshu777:
Bug :
The Progress bar indicator is shown even when there is only one slide.

Adding invisible condition for the Single Slide Progress Indicator similar to #504 and resolved similar to in a similar way #876

### DIFF
--- a/appintro/src/main/java/com/github/appintro/indicator/ProgressIndicatorController.kt
+++ b/appintro/src/main/java/com/github/appintro/indicator/ProgressIndicatorController.kt
@@ -3,6 +3,7 @@ package com.github.appintro.indicator
 import android.content.Context
 import android.graphics.PorterDuff
 import android.util.AttributeSet
+import android.view.View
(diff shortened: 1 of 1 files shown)


appintro/src/main/java/com/github/appintro/indicator/ProgressIndicatorController.kt
```diff
@@ -35,6 +36,9 @@ class ProgressIndicatorController @JvmOverloads constructor(
 
     override fun initialize(slideCount: Int) {
         this.max = slideCount
+        if (max == 1) {
```
cortinico, line 39:
```suggestion
        if (slideCount == 1) {
```

cortinico (approved):
Thanks for the contrib @Debanshu777 🙏 
Just one minor thing to change

cortinico (approved):

PR #886, opened 2020-10-23
Fixed Readme permissions example mistake

cyb3rko:
I suppose you forgot to change the second example after copy pasting, so I (hopefully) fixed it.
Is it correct?

Greetings!


cortinico:
Thank you very much for catching this 👍

PR #887, opened 2020-10-23
Added start/stop animation for `Animatable` images

sr01:


appintro/src/main/java/com/github/appintro/model/SliderPage.kt
```diff
@@ -25,6 +26,7 @@ data class SliderPage @JvmOverloads constructor(
     var title: CharSequence? = null,
     var description: CharSequence? = null,
     @DrawableRes var imageDrawable: Int = 0,
+    var startAnimation: Boolean = false,
```
cortinico, line 29: I think we should not add `startAnimation` as a boolean flag to `SliderPage` (and generally be exposed).

cortinico (changes requested):
Thanks for your contrib.

I'm fine with the idea that if the provided Drawable is an Animatable, then we can animate it.

However I would suggest we don't expose a new `startAnimation` flag.

appintro/src/main/java/com/github/appintro/model/SliderPage.kt
sr01: Agree, its not required.

cortinico:
@sr01 What's the status of this PR?

cortinico (approved):

PR #892, opened 2020-10-25
Run Pre merge checks only on master

cortinico:
I've set up two shadow CI jobs to test how Kotlin EAP and AGP alpha behaves with our library:
https://github.com/AppIntro/AppIntro/commit/817a633771733196dc376cfa451c2c778d1babd1

Here I'm cleaning up the job to make sure that we run only on master and not on all the branches.


paolorotolo (approved):
Okay, I also had a look at the rest of the code of the shadow CI jobs, looks good!

PR #895, opened 2020-10-31
Azerbaijan language support

ch22843:


cortinico (approved):
Thank you very much for the contribution @ch22843

PR #896, opened 2020-11-03
Fix for crash on PagerAdapter when using SlidePolicy

### LINKED ISSUE #893: Orientation changes make views unreachable
**AppIntro Version**:
6.0.0

**Device/Android Version**:
Samsung Galaxy S7/Android 8.0.0, Google Pixel 4 Emulator/Android 11

*UPDATE*:
As I just figured out also your example slide policy fragment does crash.

I didn't specify any orientation in the Manifest on my AppIntro actvity so orientation changes are allowed.
It works, it also changes the portrait layout to a landscape layout, but if the layout changes I can not get the binding/view of any fragment anymore. It always crashes when I want to access the view and e.g. check the state of checkboxes.

It seems like there`s something wrong in how the lifecycle is used to recreate the view.
Is there a workaround?

**Your Code / LogCat**: 
1. using single views
In my onCreateView method of my custom fragment I initialize my checkbox like so:
`checkBox1 = view.findViewById(R.id.terms_of_use_check)`
If I wanna get the state as the `isPolicyRespected` value, it crashes and says `NullPointerException at getBinding`.

2. using viewBinding (for source go [HERE](https://github.com/cyb3rko/AppIntroCrash))
In my onCreateView method of my custom fragment I initialize my binding like so:
`_binding = FragmentIntroBinding.inflate(layoutInflater, container, false)`
If I wanna get the state as the `isPolicyRespected` value, it crashes and says `NullPointerException at getBinding`.

**Example video**:
https://youtube.com/watch?v=i650NLLtR_8

cortinico:
This fixes a crash that arise when orientation changes are happening on a `SlidePolicy` Fragment.
Steps to reproduce the crash are available in the description of #893 

Fixes #893

### DIFF
--- a/appintro/src/main/java/com/github/appintro/internal/viewpager/PagerAdapter.kt
+++ b/appintro/src/main/java/com/github/appintro/internal/viewpager/PagerAdapter.kt
@@ -1,16 +1,23 @@
 package com.github.appintro.internal.viewpager
 
+import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentManager
 import androidx.fragment.app.FragmentPagerAdapter
 
 internal class PagerAdapter(
     fragmentManager: FragmentManager,
-    private val fragments: List<Fragment>
+    private val fragments: MutableList<Fragment>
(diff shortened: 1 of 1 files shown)


paolorotolo (approved):
LGTM, thanks!

Issue #865, opened 2020-09-03
Support for animated drawables


JonasBernard:
Hi! I tried implementing an onboarding screen with an animated drawable as drawable.

I'm not ready with that and my solutions seems to get very ugly.
So I thought, maybe you could add this feature.

Animatable drawables can be handled as normal drawables, but have to be started as mentioned here:
https://developer.android.com/guide/topics/graphics/drawable-animation#AnimationDrawable

cortinico:
Ideally you could just use a custom layout fragment to support `AnimatedVectorDrawable`. The solution should also be pretty straightforward.

> I'm not ready with that and my solutions seems to get very ugly.

Would you be able to share your solution? We can discuss it and see if we eventually add it as an example to the AppIntro sample app? 👌

PR #898, opened 2020-11-05
Sample for using animated drawables

cyb3rko:
Hey again,
as requested in #865 I added a pretty simple sample of how to use animated drawables using a custom layout.

I used a AnimatedDrawable by creating the animation via an animation list.
The layout size of the ImageView is to be improved because I think currently it's not that dynamical.
Maybe you know how to solve this problem,..

I hope it's okay.


example/src/main/java/com/github/appintro/example/ui/custom/fragments/AnimatedDrawableFragment.kt
```diff
@@ -0,0 +1,38 @@
...
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? = inflater.inflate(R.layout.intro_animated_drawable, container, false)
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        view.findViewById<ImageView>(R.id.animated_image_view).apply {
+            setBackgroundResource(R.drawable.animated_drawable)
```
cortinico, line 26: How about if we use instead an `AnimatedVectorDrawable`?

example/src/main/java/com/github/appintro/example/ui/custom/fragments/AnimatedDrawableFragment.kt
```diff
@@ -0,0 +1,38 @@
...
+
+class AnimatedDrawableFragment : Fragment() {
+
+    private lateinit var myAnimation: AnimationDrawable
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? = inflater.inflate(R.layout.intro_animated_drawable, container, false)
```
cortinico, line 20: `R.layout.intro_animated_drawable` is probably not needed and you should be fine using the layouts provided by AppIntro.

cortinico (commented):
Thanks for taking the time to add this 💪

example/src/main/java/com/github/appintro/example/ui/custom/fragments/AnimatedDrawableFragment.kt
cyb3rko: The problem is that the default layouts don't crop and fit the drawables correctly, you can try it with the default layouts, it looks weird.
If you don't have time, I can show you how it looks when I get back from school.

example/src/main/java/com/github/appintro/example/ui/custom/fragments/AnimatedDrawableFragment.kt
cyb3rko: Oh, I've just seen that you suggested using AnimatedVectorDrawable in the issue, sorry.
I can change it today.

example/src/main/java/com/github/appintro/example/ui/custom/fragments/AnimatedDrawableFragment.kt
cortinico: > The problem is that the default layouts don't crop and fit the drawables correctly, you can try it with the default layouts, it looks weird.

This looks like a bug honestly. If so, we should fix it outside of this PR

cyb3rko: I would not call it a bug.
I think it's just because it's an animated drawable and it does not work for animated drawables.

cyb3rko: Ok, it's not a bug. The used drawables were just too small.
Replaced the old drawables with new ones.
Also now using the default intro layout.

example/src/main/java/com/github/appintro/example/ui/custom/fragments/AnimatedDrawableFragment.kt
cyb3rko: The minimum api level is 21. Shall we then just leave it as is?

cortinico: Have you tried using `vectorDrawables.useSupportLibrary = true`
 https://developer.android.com/guide/topics/graphics/vector-drawable-resources#vector-drawables-backward-solution?

cyb3rko: Oh, didn't see, sry. I will try it out

PR #900, opened 2020-11-07
Update README.md

cyberneid:
Added another app using AppIntro: "Firmo con CIE"
https://play.google.com/store/apps/details?id=com.cyberneid.disigoncie


cortinico (approved):
Grazie 👌

PR #901, opened 2020-11-12
ISlidePolicy Example

ELIX1337:
1. The **policy** slide can be skipped by pressing the SKIP button.
2. User can restart the application during slide shows (reset the first run flag ) and thus skip the user agreement slide


cortinico:
Hey @ELIX1337, are you reporting an issue?
If so, please open an issue and not a Pull Request.

PR #903, opened 2020-11-14
Deprecate the setNextPageSwipeLock method

### LINKED ISSUE #902: Question : how do setButtonsEnabled / setNextPageSwipeLock work ?
I'm calling `setButtonsEnabled ` and `setNextPageSwipeLock` in the `onSlideChanged` method but it doesn't have any effect on navigation : buttons are still enabled and user can still swipe.

Is there something I'm doing wrong ?

**AppIntro Version**: 6.0.0

**Device/Android Version**: Huawei 8X

**Issue details / Repro steps / Use case background**: Using an activity extending `AppIntro2` with specialized fragments for certain slides

**Your Code**: 

```java

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        addSlide(new WelcomeIntroFragment());
        addSlide(new ImportIntroFragment());
        addSlide(new ThemeIntroFragment());

        setWizardMode(true); // Replaces skip button with back button
        setSystemBackButtonLocked(true); // works
        setIndicatorEnabled(true);
        setSwipeLock(true); // <--- Doesn't work
    }

 @Override
    public void onSlideChanged(@Nullable Fragment oldFragment, @Nullable Fragment newFragment) {
        super.onSlideChanged(oldFragment, newFragment);
        boolean isBlockProgress = (newFragment instanceof ImportIntroFragment);
        setButtonsEnabled(!isBlockProgress); // <-- doesn't have any effect
        setNextPageSwipeLock(isBlockProgress); // <-- doesn't have any effect
    }
```

**Stack trace / LogCat**:
```ruby
paste stack trace and/or log here
```
No error message; none expected

cortinico:
I'm deprecating the `setNextPageSwipeLock` method on `AppIntroBase`.

The method has no effect at all on the ViewPager and is causing confusion on the status of buttons if users happen to call it. We're going to remove the method in the upcoming major version of the library.

Fixes #902

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
@@ -238,22 +238,19 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
     }
 
     /**
-     * Setting to disable forward swiping right on current page and allow swiping left. If a swipe
-     * left occurs, the lock state is reset and swiping is re-enabled. (one shot disable) This also
-     * hides/shows the Next and Done buttons accordingly.
-     *
      * @param lock Set true to disable forward swiping. False to enable.
+     * @deprecated setNextPageSwipeLock has been deprecated in favor of setSwipeLock or SlidePolicy
      */
+    @Deprecated(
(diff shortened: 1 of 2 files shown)


paolorotolo (approved):
Ok for me.

PR #912, opened 2020-12-18
Use fragments instead of activities

nasko700:
It could be created another module of this nice library which uses fragments instead of activities. Here AppIntro is changed to be a fragment.


cortinico:
Thanks for your contribution @nasko700 

What should be the going forward here? This PR is introducing a drastic API change with no backward compatibility with AppIntro 6.0.0.

I appreciate the effort but I believe this should be split in smaller parts and the library should expose both `Activity` and `Fragment` support.

PR #922, opened 2021-01-28
Add support for Autosizing TextViews

### LINKED ISSUE #915: A display issue
**AppIntro Version**: 2.15.15

**Device/Android Version**: Google nexus 6，Android 8.0

Dear developer!
In the process of using, I found a bug in layout / display. As shown in the figure, in the first page, the text in the red box is half blocked. This may be an adaptation problem. I suspect it is because **display and font size** are set to large (mainly due to **display size**).
Because of some problems with my eyesight, I often use the largest font and display.
App version is 2.15.15, my mobile phone is Google nexus 6, screen resolution setting is 1920* 1080, **_font size and display size are the largest ! ! !_**, Android system is 8.0.
The above is a description of the problem. The overall experience of the software is good. You can see that the design is very attentive. Thank you very much for the efforts of the developers all the time. I hope to receive a reply.
Thank you!

**Your Code**: 

**Stack trace / LogCat**:
```ruby
paste stack trace and/or log here
```

bug-beta_342

cortinico:
Fixes #915 

I'm adding Autosizing TextViews (https://developer.android.com/guide/topics/ui/look-and-feel/autosizing-textview) to fix problems with text being cut out.

| Long text | Short test |
| --- | --- |
| Screenshot_1611874511 | Screenshot_1611874513 |

### DIFF
--- a/appintro/src/main/res/values-sw600dp/dimen-sw600dp.xml
+++ b/appintro/src/main/res/values-sw600dp/dimen-sw600dp.xml
@@ -3,6 +3,8 @@
 
     <dimen name="appintro_desctext_size">30sp</dimen>
     <dimen name="appintro_headtext_size">42sp</dimen>
+    <dimen name="appintro_min_text_size">24sp</dimen>
+    <dimen name="appintro_text_granularity_size">2sp</dimen>
 
     <dimen name="appintro_button_margin">12dp</dimen>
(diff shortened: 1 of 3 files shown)


paolorotolo (approved):
Awesome, I didn't know this was a thing 👍🏻

PR #923, opened 2021-01-28
Refactor project to use maven-publish

cortinico:
I'm updating the project to use `maven-publish` as suggested by Jitpack: https://jitpack.io/docs/ANDROID/

We also had quite and old configuration so I'm cleaning it up a bit.


paolorotolo (approved):
Yeah, thanks!

PR #934, opened 2021-03-12
Fix ProgressIndicatorController in RTL

### LINKED ISSUE #932: RTL bug - progress indicators do not adapt to RTL
**AppIntro Version**: 6.1.0

**Device/Android Version**: Sony, Android 7.0

**Issue details**: In issue #721 you made the demo with dot indicators, I tried it with progress indicator and it is not working correctly.

cortinico:
Seems like we have a problem with `ProgressIndicatorController` if in RTL (it goes in the opposite way).
I'm fixing it here:

untitled

Fixes #932


paolorotolo (approved):
Good, thanks!

PR #948, opened 2021-05-26
Fix #926: Add color resource parameters

### LINKED ISSUE #926: Changing light/dark mode not works for color resources
**AppIntro Version**: current master (SNAPSHOT)

**Device/Android Version**: Emulator / Android 11 (API level 30)

If the intro is open and the dark mode configuration changes not all colors are changed.
1. Open intro in light mode
2. Change theme to dark mode by control center

The drawables (set by resource ID) will be changed like expected. All background & font colors (that set directly) does not change.

After the design mode changed the activity will recreated. The slides will added again with new (dark mode) parameters. The colors are set directly and I found that the onActivityCreated [AppIntroBaseFragment.kt#L73](https://github.com/AppIntro/AppIntro/blob/master/appintro/src/main/java/com/github/appintro/AppIntroBaseFragment.kt#L73) method is called while recreating with the "old" parameters and the explicit set colors will be reused.

The simplest solution is to give a color resource ID instead of the color itself. As more cleaner solution we should improve the instance saving flow to keep them per fragment itself instead to referencing to the activity lifecycle.

**Your Code**: 
```kotlin
override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)

    // Have default + night variants
    // Works fine
    setIndicatorColor(
            selectedIndicatorColor = ContextCompat.getColor(this, R.color.intro_pagination_indicator),
            unselectedIndicatorColor = ContextCompat.getColor(this, R.color.meta_grey)
    )

    // Have default + night variants
    // Works fine
    setBackArrowColor(ContextCompat.getColor(this, R.color.accent))
    setNextArrowColor(ContextCompat.getColor(this, R.color.accent))
    setColorDoneText(ContextCompat.getColor(this, R.color.paper))

    addSlide(AppIntroFragment.newInstance(
            // ...

            // Has default + night variant
            // Works fine
            imageDrawable = R.drawable.ic_logo, 

            // Has default + night variant
            // Does NOT work
            titleColor = ContextCompat.getColor(this, R.color.text),

            // Has default + night variant
            // Does NOT work
            descriptionColor = ContextCompat.getColor(this, R.color.text),

            // Has default + night variant
            // Does NOT work
            backgroundColor = ContextCompat.getColor(this, R.color.paper),
            
            // ...
    ))
}
```

Working light:

Failed dark:

Expected result:

I will playing around a little bit and make a PR next days.

Fabi755:
As we discussed here the PR for #926. I had added 3 new resource parameters for color definitions.

While booth parameter types are from type Integer, I decided to create a `newInstance2(...)` method. I think this is a more cleaner solution while Kotlin does not support deprecation marks on function parameters.

I added deprecation marks for the existing values. But is that the right choice? 
On NavBar and StatusBar color function the library does support both options static color and a resource value:
> // Control the status bar color
> setStatusBarColor(Color.GREEN)
> setStatusBarColorRes(R.color.green)
> 
> // Control the navigation bar color
> setNavBarColor(Color.RED)
> setNavBarColorRes(R.color.red)

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,11 +81,11 @@ class MyCustomAppIntro : AppIntro() {
 
         // Call addSlide passing your Fragments.
         // You can use AppIntroFragment to use a pre-built fragment
-        addSlide(AppIntroFragment.newInstance2(
+        addSlide(AppIntroFragment.newInstanceWithRes(
                 title = "Welcome...",
                 description = "This is the first slide of the example"
         ))
-        addSlide(AppIntroFragment.newInstance2(
+        addSlide(AppIntroFragment.newInstanceWithRes(
                 title = "...Let's get started!",
                 description = "This is the last slide, I won't annoy you more :)"
         ))
@@ -150,7 +150,7 @@ You can use the `AppIntroFragment` if you just want to customize title, descript
 That's the suggested approach if you want to create a quick intro:
 
 ```kotlin
-addSlide(AppIntroFragment.newInstance2(
+addSlide(AppIntroFragment.newInstanceWithRes(
     title = "The title of your slide",
     description = "A description that will be shown on the bottom",
     imageDrawable = R.drawable.the_central_icon,
--- a/README.md
+++ b/README.md
@@ -81,11 +81,11 @@ class MyCustomAppIntro : AppIntro() {
 
         // Call addSlide passing your Fragments.
         // You can use AppIntroFragment to use a pre-built fragment
-        addSlide(AppIntroFragment.newInstance(
+        addSlide(AppIntroFragment.newInstance2(
                 title = "Welcome...",
                 description = "This is the first slide of the example"
         ))
-        addSlide(AppIntroFragment.newInstance(
+        addSlide(AppIntroFragment.newInstance2(
                 title = "...Let's get started!",
                 description = "This is the last slide, I won't annoy you more :)"
         ))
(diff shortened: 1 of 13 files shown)


cortinico:
Thanks for doing this 🙏 

> While booth parameter types are from type Integer, I decided to create a `newInstance2(...)` method. I think this is a more cleaner solution while Kotlin does not support deprecation marks on function parameters.

Indeed 👍  But `newInstance2` is not the best name. I would suggest we rename it to `newInstanceWithRes` or `newInstanceRes`.

I was also thinking if we should instead just add 3 `Int` parameters to `newInstance` without creating a new method, but I think the solution will be not as clean.

Fabi755:
I have changed the naming 👍 

I think if we decide against the deprecation and want to support both color resource types, we can migrate the parameters back into one `newInstance` function. But with deprecation we should keep the second function until we have removed the old stuff.

cortinico (approved):

cortinico:
Sorry for the really late review.
I believe we can rebase this on top of #958 and merge it just after 👍

cortinico:
@Fabi755 Can you please rebase or give edit permission to contributors so I can land this PR?

tmcintosh:
Would be great to get his merged!  Thanks for the work @Fabi755

Fabi755:
I will rebase this later today, sorry for the delay.. holiday time is now over ;-)

cortinico:
What's the status of this PR? I'd like to include it inside AppIntro 6.2 eventually. @paolorotolo could you also take a quick look?

Fabi755:
> What's the status of this PR? 

All things are done. Last changes of this PR was the rebase to master branch. At least tests needs to be run & a final review must be made 😄

cortinico:
> At least tests needs to be run & a final review must be made 😄

I don't know why but the CI is not running. My hope is that the rebase would have triggered it but it's not the case apparently.
I would suggest that you either enable maintainer access to your PR or you close this and reopen a new PR. Hopefully Github Actions will run then 👍

Fabi755:
I think it's cause by my mistake that I used forced push to got the rebase done.

I will re-create this PR to get a clean state. Sorry for that!

PR #959, opened 2021-08-06
Add @JvmOverloads on goToNextSlide

### LINKED ISSUE #955: Calling goToNextSlide() from Java
## ⚠️ Is your feature request related to a problem? Please describe

goToNextSlide() has default parameter values however it's impossible to use goToNextSlide from Java because it requires isLastSlide which relies on internal logic.

From Kotlin document:

> Normally, if you write a Kotlin method with default parameter values, it will be visible in Java only as a full signature, with all parameters present. If you wish to expose multiple overloads to Java callers, you can use the @JvmOverloads annotation.

Please add @JvmOverloads annotation for goToNextSlide so we can use this method from Java without the need to provide isLastSlide.

## 🤚 Do you want to develop this feature yourself?
- [ ] Yes
- [X] No

cortinico:
Fixes #955

Currently `goToNextSlide` is not callable from Java because the parameter `isLastSlide` relies on internal logic. Adding `@JvmOverloads` will solve this.

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
@@ -174,6 +174,7 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
     }
 
     /** Moves the AppIntro to the next slide */
+    @JvmOverloads
     protected fun goToNextSlide(isLastSlide: Boolean = pager.isLastSlide(fragments.size)) {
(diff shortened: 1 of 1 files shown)


cortinico:
Lint is failing due to dependencies not up to date. #953 is fixing this.

paolorotolo:
Nice, thanks.

cortinico:
You can follow the update for 6.2 here https://github.com/AppIntro/AppIntro/pull/978

PR #962, opened 2021-08-07
Java example: Use setImmersiveMode instead of setImmersive

cortinico:
I accidentally merged this in the Java example. This method calls should be removed as it requires API lvl 18 and is making lint unhappy.


paolorotolo (changes requested):
I'd say, call AppIntro's `setImmersiveMode` instead of Activity's `setImmersive` that doesn't require higher API levels and offers the same behaviour

cortinico:
> I'd say, call AppIntro's `setImmersiveMode` instead of Activity's `setImmersive` that doesn't require higher API levels and offers the same behaviour

Right :) Updated it

paolorotolo (approved):
Awesome, ty!

Issue #963, opened 2021-08-28
Descriptions are blocked when using the largest font and display size


yuhuitech:
## 🐛 Describe the bug
Hello! Dear developer!
In the process of using, I found a bug in layout / display. As shown in the figure below, on those intro pages, **there are some texts that could not be seen fully, and they are not scrollable as well**. This may be an adaptation problem. I suspect it is because **display and font size** are set to large **(mainly due to display size)**.
Because of some problems with my eyesight, I often use the largest font and display.
![103165857-dbaf8580-4857-11eb-9e22-120ae1901389](https://user-images.githubusercontent.com/25502419/128597553-7949a1a3-0b21-4ddf-9357-766b5c302518.png)

## ⚠️ Current behavior
There are some texts that could not be seen fully when I set the display size and font to the largest, especially when the description is very long, which might overflow and there is no way for me to see them, unless I shrink the size. And as far as I have known, there are a lot of other apps that applied your package and also had this problem. I would appreciate it if you could review my issue. :)

## ✅ Expected behavior
I hope to see those descriptions fully.

## 💣 Steps to reproduce
1. use the largest font and display in the display setting
2. open this app(AppIntro Sample App), then click the start button on the Default Intro card.
3. this error could be seen when you scroll all the way down to the third page.

## 📷 Screenshots
The blocked part has been circled in red already.

<img src="https://user-images.githubusercontent.com/25502419/131211862-100bdb96-3e67-4970-81e2-148625ffa2b3.png" alt="copy" width="250"/>

## 📑 Your Code
I would like to solve this problem with a pull request! :)

## 📱 Tech info
 - AppIntro Version: 6.1.0
 - Device: Pixel 2
 - Android OS Version: 10

cortinico:
Thanks for your report. I've followed up on #964

PR #964, opened 2021-08-28
Make the descriptions on the intro pages scrollable

yuhuitech:
Dear developer:

Hello! I am the creator of issue #963, and I have fixed the blocked descriptions to make it more responsive. I would appreciate it if you could kindly revise my code and leave me some advice. Thank you so much for your precious time!

Here are the screenshots after my changes:

copy copy copy

Thanks again! Blessings on your day! :)

### DIFF
--- a/appintro/src/main/res/layout/appintro_fragment_intro.xml
+++ b/appintro/src/main/res/layout/appintro_fragment_intro.xml
@@ -30,22 +30,27 @@
         android:layout_height="0dp"
         app:layout_constraintVertical_weight="5"
         android:contentDescription="@string/app_intro_image_content_description"
-        app:layout_constraintBottom_toTopOf="@+id/description"
+        app:layout_constraintBottom_toTopOf="@+id/scrollable_textview"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@+id/title" />
 
-    <TextView
-        android:id="@+id/description"
-        style="@style/AppIntroDefaultText"
+    <ScrollView
         android:layout_width="0dp"
         android:layout_height="0dp"
-        android:autoLink="web"
+        android:id="@+id/scrollable_textview"
         app:layout_constraintVertical_weight="3"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@+id/image"
-        tools:text="This is a demo of the AppIntro Library" />
+        app:layout_constraintTop_toBottomOf="@+id/image">
+        <TextView
+            android:id="@+id/description"
+            style="@style/AppIntroDefaultText"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:autoLink="web"
+            tools:text="This is a demo of the AppIntro Library" />
+    </ScrollView>
 
 </androidx.constraintlayout.widget.ConstraintLayout>
--- a/appintro/src/main/res/layout-land/appintro_fragment_intro.xml
+++ b/appintro/src/main/res/layout-land/appintro_fragment_intro.xml
@@ -12,11 +12,14 @@
     <TextView
         android:id="@+id/title"
         style="@style/AppIntroDefaultHeading"
+        app:layout_constraintVertical_weight="1"
+        app:layout_constraintHorizontal_weight="1"
+        android:gravity="center"
         android:layout_width="0dp"
-        android:layout_height="wrap_content"
-        app:layout_constraintBottom_toTopOf="@+id/description"
+        android:layout_height="0dp"
+        app:layout_constraintBottom_toTopOf="@+id/landing_scroll_description"
         app:layout_constraintEnd_toEndOf="parent"
-        app:layout_constraintStart_toStartOf="@+id/midline"
+        app:layout_constraintStart_toEndOf="@+id/image"
         app:layout_constraintTop_toTopOf="parent"
         app:layout_constraintVertical_chainStyle="spread"
         tools:text="Welcome" />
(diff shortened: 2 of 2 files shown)


franklinbill:
I met this mistake again. Great! I agree with your fix 👍 @yuhuitech

appintro/src/main/res/layout/appintro_fragment_intro.xml
```diff
@@ -30,22 +30,27 @@
...
-        app:layout_constraintBottom_toTopOf="@+id/description"
+        app:layout_constraintBottom_toTopOf="@+id/scrollable_textview"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@+id/title" />
 
-    <TextView
-        android:id="@+id/description"
-        style="@style/AppIntroDefaultText"
+    <ScrollView
```
cortinico, line 38: The scrollview is not needed. It should be enought with: 
```
android:scrollbars = "vertical"
```

Plus 

```java
textView.setMovementMethod(new ScrollingMovementMethod());
```

cortinico (commented):
Thanks for your PR 🙏 

The `layout-land` file should also be adapted.

appintro/src/main/res/layout/appintro_fragment_intro.xml
yuhuitech: OK, I will work on this right now. :) 
Thank you so much!

yuhuitech: Hi! Those two methods seem to have the same effect as far as I am concerned. And I also guess it would be easier to maintain the project if we let the XML do all the rending work. Editing Java files would also work, but maybe it is harder for other developers to associate it with the XML file. This is my opinion. I would appreciate it if you could give me some advice on this. Thank you so much! :)

yuhuitech: If it is OK with you I would like to do the adaptation work of the land mode. :)

cortinico: Let's go with the XML only approach 👍

cortinico: Please note this: 

> The scrollview is not needed.

You can remove the ScrollView entirely 👍

yuhuitech: Hi! Sorry for the wait. I guess I have solved the problem when the landscape mode is used. Here are some screenshots after modifications when using the largest display size:

copy
copy 
copy

and the smallest display size:

copy
copy

I am looking forward to your review. Thanks again for your kindness and your precious time!! :)

yuhuitech: `You can remove the ScrollView entirely`

If I remove the ScrollView, should I use the java code you recommended earlier?

cortinico: Please do your own researches/trial and let me know what's the outcome.
The solution with the least amount of changes is generally preferred.

Like if we can solve it by just adding an XML attribute, that would be the preferred approach.

yuhuitech: OK, I could comprehend the meaning you are trying to deliver to me now. Therefore I would like to make a comparison of those two methods and I would like to inform you of this once it is ready. And I will also minimize the changes I could apply to those files. :)
Thanks again!

yuhuitech:
Hello! Sorry for the wait. After my research on this project and this problem, I would like to recommend you modify XML files only. Here is the reason.

I did some investigations on this project and I found that maybe it is the "AppIntroBaseFragment.kt" file that controls the creation of the views on the Intro pages, which is this function `onCreateView`, more specifically. I applied some changes to it, just like this:
```
    override fun onCreateView(
        inflater: LayoutInflater,
        container: ViewGroup?,
        savedInstanceState: Bundle?
    ): View? {
        val view = inflater.inflate(layoutId, container, false)
        val titleText = view.findViewById<TextView>(R.id.title)
        val descriptionText = view.findViewById<TextView>(R.id.description)
        val slideImage = view.findViewById<ImageView>(R.id.image)
        val mainLayout = view.findViewById<ConstraintLayout>(R.id.main)

        // Make the description texts scrollable
        descriptionText.setMovementMethod(ScrollingMovementMethod())

        titleText.text = title
        descriptionText.text = description
        if (titleColor != 0) {
            titleText.setTextColor(titleColor)
        }
        if (descColor != 0) {
            descriptionText.setTextColor(descColor)
        }
        titleTypeface?.applyTo(titleText)
        descTypeface?.applyTo(descriptionText)

        slideImage.setImageResource(drawable)
        if (bgDrawable != 0) {
            mainLayout?.setBackgroundResource(bgDrawable)
        } else {
            mainLayout?.setBackgroundColor(defaultBackgroundColor)
        }

        return view
    }
```
In the meantime, I also revoked all the changes I applied to those two XML files. Then I started my test. 

Here are some normal screenshots after my test, and in such circumstances, those text views could be seen fully:

copy copy 

And it also could handle the situation where an extremly longer sentence is in the text view.

copy

However, this screenshot is not that charming when it comes to the land mode, mainly for the title. As you could see, a part of the title is blocked by its border. But the description is scrollable.

copy

According to my comparison, I found some layout issues in the land mode, which are caused by the XML file. So I guess the land mode XML file needs a repair. From the screenshots I have shown you before, I reckon that I have tackled this problem without changing any kt or java code, just some modifications on the XML files. That is, after my modifications, the long title on the land mode is not blocked.

So my final conclusion is that perhaps we don't need to change any kt or java files but only XML files to reach this goal.

That is my opinion. I am very humble and sincere so I would be so glad if you could leave me some advice on this. And I would like to make contributions to this project with all of my best efforts. :)

Thanks again!

cortinico:
Thanks for your clarification and for taking the time to look into this,

> So my final conclusion is that perhaps we don't need to change any kt or java files but only XML files to reach this goal.

Are you suggesting we go with the `ScrollView` approach right?

yuhuitech:
Yes, we should go with the `ScrollView` approach. :)

cortinico:
> Yes, we should go with the `ScrollView` approach. :)

I would like to take a view at my implementation here: https://github.com/AppIntro/AppIntro/commit/91909ec72bfe8e558942f598a2c9da79a2135b0d

It doesn't use the `ScrollView` yet achieves the same solution.
I believe the only problem left is the Title in Land mode, but I think we should tune the text autosizing for that specific scenario.

yuhuitech:
Yes, I also believe tuning is necessary and I would like to try your code.

cortinico:
> Yes, I also believe tuning is necessary and I would like to try your code.

What's the status of this PR?

yuhuitech:
> > Yes, I also believe tuning is necessary and I would like to try your code.
> 
> What's the status of this PR?

Hello! I am quite busy recently, and I have tested your code already, and I think they are good to go. Making those text scrollable is a wonderful choice.

cortinico:
Thanks for the feedback. I believe we might want to include this change into #978

PR #970, opened 2021-10-03
Fix java example link in readme

Starlight220:


cortinico (approved):
Thank you for the fix 🙏

PR #979, opened 2021-11-01
Fix #926: Add color resource parameters

### LINKED ISSUE #926: Changing light/dark mode not works for color resources

Fabi755:
Fixing #926. As discussed in #948 this PR is re-created for a cleaner state.


paolorotolo (commented):
Sorry for writing this only now.
The code looks fine for me, but I'm just unsure about the naming of the new method `newInstanceWithRes`. 

It's possible to invoke that method passing the `title` and `description` and leaving the default values of the other parameters.
```
        addSlide(AppIntroFragment.newInstanceWithRes(
                "Welcome!",
                "This is a demo of the AppIntro library, with a custom background on each slide!",
        ))
```
In that case, I think that the naming `newInstaceWithRes` would be misleading, because the user expects that a resource is required by the method.

Since this will be the default method to generate instance of AppIntroFragment, I think we should find something more neutral in that sense.

cortinico:
> In that case, I think that the naming `newInstaceWithRes` would be misleading, because the user expects that a resource is required by the method.
> 
> Since this will be the default method to generate instance of AppIntroFragment, I think we should find something more neutral in that sense.

How do we feel about `createInstance()`?

paolorotolo:
Yes, sounds good for me 👍🏻

paolorotolo:
> How do we feel about `createInstance()`?

Yes, sounds good to me! Could you please refactor that?

Fabi755:
So I do the following points:

- [ ] rename `newInstaceWithRes` to `createInstance`
- [ ] solve conflict

Is there something more?

appintro/src/main/java/com/github/appintro/AppIntroFragment.kt
```diff
@@ -32,6 +33,15 @@ class AppIntroFragment : AppIntroBaseFragment() {
          */
         @JvmOverloads
         @JvmStatic
+        @Deprecated(
+            "`newInstance` is deprecated to support color resources with configuration " +
```
cortinico, line 37:
```suggestion
            "`newInstance` is deprecated to support color resources instead of color int for configuration changes and dark theme support "
```

Let's try to be a bit more informative

cortinico:
> Is there something more?

Left a minor comment but otherwise looks great to me 👍

cortinico (approved):

paolorotolo:
Thanks!

PR #980, opened 2021-11-05
Rename branch references from master to main

cortinico:
As Github suggested, I've renamed the default branch from `master` to `main`. Here I'm renaming the references we had of master in the codebase.


paolorotolo:
Thanks, lint check should be fixed after merging #981

PR #981, opened 2021-11-05
Bump androidx annotations to 1.3.0

cortinico:
Just noticed that Lint was failing because of this. I'm bumping this dependency.


paolorotolo:
I'm ignoring the check with AGP-7.2.+ since it just requires a higher version of gradle.

paolorotolo (approved):

PR #984, opened 2021-11-07
Fix wrong method name in deprecated annotation

paolorotolo:


paolorotolo:
Rebased on main to include lint fix

cortinico (approved):

PR #997, opened 2022-01-26
Set pager offscreenPageLimit for all modes

### LINKED ISSUE #995: Load issue after rotate two times
## 🐛 Describe the bug
Rotated two times and then swiped pages and went to the third page, but it didn't load immediately.
When I set transformer to AppIntroPageTransformerType.Fade, It didn't load anymore. I should go to the next slide and then back to that to load that page.

## ⚠️ Current behavior
Third fragment loads after a second with the default transformer
Third fragment doesn't load with the Fade transformer

## ✅ Expected behavior
It should load immediately like other fragments

## 💣 Steps to reproduce
1 - Go to Default Intro
2 - Rotate device to Landscape
3 - Rotate device to Portrait
4 - Swipe and go to the second page
5 - Swipe and go to the third page

## 📷 Screenshots
https://user-images.githubusercontent.com/8654398/150964255-b6d82547-20f0-48c7-a475-fac45710aad2.mp4

## 📑 Your Code
example code (default intro)

## 📱 Tech info
 - AppIntro Version: 6.2.0
 - Branch: master
 - Head commit: 65981e52426a21fb3fdf419e6bbcf86e5c8c685b
 - Device: Android Emulator
 - Android OS Version: 11

rezazarchi:
resolves #995

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
@@ -149,9 +149,7 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
         } else {
             fragments.add(fragment)
         }
-        if (isWizardMode) {
-            pager.offscreenPageLimit = fragments.size
-        }
+        pager.offscreenPageLimit = fragments.size
         pagerAdapter.notifyDataSetChanged()
     }
 


cortinico:
Hey @rezazarchi 
I've just tested the original bug reported in #995 and I wasn't able to reproduce at all. 
I've tried with the sample app and several of the sample AppIntro from there.
Can you clarify how to reproduce it?

rezazarchi:
Sorry. this happens only on RTL layouts.
You should enable force RTL layout direction in Developer options or choose a RTL language from your device language settings (like Arabic or Persian)

cortinico:
Sorry this took a bit longer than expected. I believe the correct way to fix this is the one I pushed on #1002. Setting the `offscreenPageLimit` is just a patch but is not fixing the underlying problem. Let me know what you think about it.

rezazarchi:
I checked #1002 and it resolved the problem. Thanks.

PR #1002, opened 2022-02-18
Fix RTL bug on wrongly retained currentItem

### LINKED ISSUE #995: Load issue after rotate two times

cortinico:
Fixes #995
Closes #997 

This PR is fixing a bug with the RTL and the retention of the ViewPager. I believe this is the root cause of #995. Maybe @rezazarchi can verify that my patch works?

Essentially AppIntro was not correctly saving the current item when on RTL. So rotating the device when on RTL would cause to jump between slides. This PR fixes it.

| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
@@ -486,7 +486,13 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
             putBoolean(ARG_BUNDLE_IS_SKIP_BUTTON_ENABLED, isSkipButtonEnabled)
             putBoolean(ARG_BUNDLE_IS_INDICATOR_ENABLED, isIndicatorEnabled)
 
-            putInt(ARG_BUNDLE_CURRENT_ITEM, pager.currentItem)
+            // We can't use pager.currentItem as we need the current item that is RTL-invariant.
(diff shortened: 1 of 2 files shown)


paolorotolo (approved):
Looks good for me. @rezazarchi also confirmed [here](https://github.com/AppIntro/AppIntro/pull/997) that this is fixing the issue. Thanks!

PR #1025, opened 2022-09-13
Android 13 support, Kotlin 1.7.x, AppCompat 1.5.x and bump of several depedencies

cortinico:
I'm doing a general refresh of the project to make it work with Android 13:
- Bumped `compileSdk` to 33.
- Test that everything works on Android 13. The only things to tweak was the scoped storage permission in the sample app that I updated.
- Bumped Kotlin to 1.7.10
- Bumped AppCompat to 1.5.x
- Bumped Detekt to 1.21.x
- Bumped KtLint Gradle to 11.0.0
- Bumped Mockito Core to 4.8.0
- Bumped Gradle to 7.5.1
- I've cleaned up a bit of the Gradle setup as we had some stale configs.
- I've removed the experimental strategy as we're not keeping it much up to date, so we're getting little value from it.


example/build.gradle
```diff
@@ -2,31 +2,18 @@ apply plugin: 'com.android.application'
...
 android {
-    compileSdkVersion 31
+    compileSdk 33
 
     defaultConfig {
         applicationId "com.github.appintro.example"
-        minSdkVersion 14
-        targetSdkVersion 31
-        versionCode 2
-        versionName "1.0.1"
```
paolorotolo, line 12: Why are we removing those?

appintro/build.gradle
```diff
@@ -5,17 +5,14 @@ apply plugin: 'io.gitlab.arturbosch.detekt'
...
 version = "6.2.0"
 
 android {
-    compileSdkVersion 31
+    compileSdk 33
 
     defaultConfig {
-        minSdkVersion 14
-        targetSdkVersion 31
-        versionCode 20
```
paolorotolo, line 17: Same here

paolorotolo (commented):
Thanks, all looks good and works as expected. Just asking about versions removed in build.gradle (they're also missing after kts rewrite in https://github.com/AppIntro/AppIntro/pull/1030/)

appintro/build.gradle
cortinico: As they're not really needed for libraries: https://developer.android.com/studio/publish/versioning

example/build.gradle
cortinico: As we're not publishing this App (are we?).

paolorotolo (approved):

PR #1030, opened 2022-09-13
Gradle KTS & Versions Catalog

cortinico:
This PR sits on top of #1025

I've moved the whole project to use Gradle KTS and Version Catalog. It should be easier for renovate to pickup changes with a single file containing all the versions.


paolorotolo (approved):
Awesome! Looks good for me, just solve some conflicts after #1025 merge. It should be enough to accept our .kts versions of the files.

PR #1040, opened 2022-09-17
Handle deprecation of .vibrate()

cortinico:
The `.vibrate(long)` is deprecated in favor of
`.vibrate(VibrationEffect)`. I'm handling it here while still keeping the backward compatibility.

Also the `getSystemService(VIBRATOR_SERVICE)` is deprecated, I've provided the correct implementation for this as well.

### DIFF
--- a/example/src/main/java/com/github/appintro/example/ui/java/JavaIntro.java
+++ b/example/src/main/java/com/github/appintro/example/ui/java/JavaIntro.java
@@ -56,9 +56,6 @@ public class JavaIntro extends AppIntro {
         //Activate wizard mode (Some aesthetic changes)
         setWizardMode(true);
         
-        //Show/hide skip button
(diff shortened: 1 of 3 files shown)


example/src/main/java/com/github/appintro/example/ui/java/JavaIntro.java
```diff
@@ -56,9 +56,6 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
         //Activate wizard mode (Some aesthetic changes)
         setWizardMode(true);
         
-        //Show/hide skip button
-        setSkipButtonEnabled(true);
```
cortinico, line 60: I've removed this as it was clashing with `setWizardMode(true)` resulting in having both skip and the arrow on top of each other

paolorotolo (approved):

PR #1048, opened 2022-10-01
Fix several compilation warnings

cortinico:
Another round of warnings that we can get rid of:
- `getSerializable(String)` was deprecated in API 33, now we need to use `getSerializable(String, Class)`.
- `defaultBackgroundColor` was overriding a deprecated API without re-declaring the deprecation
- `setColorFilter(Int, PorterDuff.Mode)` is deprecated in favor of `setColorFilter(ColorFilter)`. I've used a compat method from Androidx here to solve this deprecation.

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
@@ -244,7 +244,7 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
         ReplaceWith("setSwipeLock"),
         DeprecationLevel.ERROR
     )
-    @Suppress("UnusedPrivateMember")
+    @Suppress("UnusedPrivateMember", "UNUSED_PARAMETER")
     protected fun setNextPageSwipeLock(lock: Boolean) {
         LogHelper.w(
             TAG,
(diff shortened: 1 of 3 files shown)


appintro/src/main/java/com/github/appintro/AppIntroBase.kt
```diff
@@ -452,18 +452,10 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
...
-                        .getItem(pager.currentItem)
-                )
-            } else {
-                // Close the intro if there are no slides to show
-                finish()
-            }
+            dispatchSlideChangedCallbacks(
+                null,
+                pagerAdapter.getItem(pager.currentItem)
+            )
```
paolorotolo, line 458: It seems we don't gracefully handle intros created with 0 slides: in this case, `getItem` will access an out-of-bound element. I thought there was a regression here but the old code was crashing too (there was a useless nullability check since `getItem` was throwing an exception and not a nullable fragment).

I think we can solve this by checking if the current item is in PagerAdapter's bound and, in this case, keep the original behavior (close the intro).

```suggestion
            if (pager.currentItem < pagerAdapter.count) {
                dispatchSlideChangedCallbacks(
                    null,
                    pagerAdapter.getItem(pager.currentItem)
                )
            } else {
                // Close the intro if there are no slides to show
                finish()
            }
```

paolorotolo (changes requested):

appintro/src/main/java/com/github/appintro/AppIntroBase.kt
cortinico: Yup good call 👍

PR #1080, opened 2023-04-12
Add ability to change done button bg color

kojofosu:


appintro/src/main/java/com/github/appintro/AppIntro2.kt
```diff
@@ -81,4 +81,13 @@ abstract class AppIntro2 : AppIntroBase() {
         val skip = findViewById<ImageButton>(R.id.skip)
         skip.setColorFilter(colorSkipButton)
     }
+
+    /**
+     * Override done button drawable resource
+     *
+     * @param resource your drawable resource*/
+    fun setDoneResource (@DrawableRes resource: Int) {
```
cortinico, line 89: I guess we'd rather do something similar to `fun setImageSkipButton(imageSkipButton: Drawable) {` so this should be:

```
fun setImageDoneButton(imageDoneButton: Drawable) {
```

cortinico (changes requested):

appintro/src/main/java/com/github/appintro/AppIntro2.kt
kojofosu: @cortinico thanks..made the suggested changes

cortinico (approved):

PR #1108, opened 2023-07-11
Fix RTL detection

### LINKED ISSUE #1005: RTL layout issue after changing the app language manually
## 🐛 Describe the bug

When the default language of the system is Persian (RTL), the intro is displayed without any problems and correctly. But when the language of the system is not Persian (e.g. English) and we change the language of the app to Persian manually, the intro is still LTR and the slide shows the beginning and the end incorrectly. If you pay attention to the screenshots, you will understand better. I have tried all the old and new methods but in all of them this problem persists. Of course, there are no problems with other activities and they are displayed correctly.

## 💣 Steps to reproduce
Adding this code in activity:
```kotlin
init {
    val locale = Locale("fa", "IR")
    val configuration = Configuration()
    Locale.setDefault(locale)
    configuration.setLocale(locale)
    this.applyOverrideConfiguration(configuration)
}
```
> This code works properly in newer versions of Android, but I does not tested it on Android N and older.

Of course, I tried many other codes, and this was the simplest code available.

## 📷 Screenshots

This screenshot is when the language of the system itself is Persian:

But this screenshot is when the system language is English and I manually changed the app language to Persian

## 📱 Tech info
 - AppIntro Version: 6.2.0
 - Device: Samsung Galaxy A51 5G
 - Android OS Version: 11 R

Fabi755:
Fixes #1005

Using the activity/view context will solve the problem. Thanks to @HashirLabs!

This is not really testable. Make it sense to add RTL/LTR samples to the example app?

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
@@ -132,7 +132,7 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
     private val argbEvaluator = ArgbEvaluator()
 
     internal val isRtl: Boolean
-        get() = LayoutUtil.isRtl(applicationContext)
+        get() = LayoutUtil.isRtl(this)
 
     /*
      PUBLIC API


cortinico:
Thanks for the fix @Fabi755

PR #1109, opened 2023-07-13
Don't crash on Parallax effect

### LINKED ISSUE #1102: Error when using Parallax Effect
## 🐛 Describe the bug
When using the Parallax Transformer (Java) the app crashes immediately. It works with the Fade effect for example. Its because I am using custom slide Fragments and layouts I guess.

## 📷 Screenshots
image

## 📑 Your Code
setTransformer(new AppIntroPageTransformerType.Parallax());
in the Activities onCreate after adding the slides, also tried it before the slides

## 📱 Tech info
 - AppIntro Version:  6.1.0 
 - Device: Pixel 5 Api 31 emulated
 - Android OS Version: Android 12

cortinico:
AppIntro should not crash on custom layout if the views that needs to be animated can't be found.

Fixes #1102

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntroPageTransformerType.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroPageTransformerType.kt
@@ -1,5 +1,7 @@
 package com.github.appintro
 
+import androidx.annotation.IdRes
+
 /**
  * Sealed class to represent all the possible Page Transformers
  * offered by AppIntro.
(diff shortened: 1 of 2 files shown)


paolorotolo:
Thanks! Since this is really error-prone when users are working with custom layout, what do you think about asking users the ids of the views to animate, just for Parallax effect?

```kotlin
    class Parallax(
        val titleParallaxFactor: Double = 1.0,
        val imageParallaxFactor: Double = -1.0,
        val descriptionParallaxFactor: Double = 2.0,
        @IdRes val titleViewId: Int = R.id.title,
        @IdRes val imageViewId: Int = R.id.image,
        @IdRes val descriptionViewId: Int = R.id.description
    ) : AppIntroPageTransformerType()
```

I'm not sure about default parameters though, we could either:
 * Always require view Ids (looks weird for non custom layouts, but safer);
 * Update javadoc and README with a warning to provide appropriate ids when using custom layouts; In this case, if an Id is not found, we could log an error in console with the suggested fix instead of failing silently.

cortinico:
Great suggestions!

> * Update javadoc and README with a warning to provide appropriate ids when using custom layouts; In this case, if an Id is not found, we could log an error in console with the suggested fix instead of failing silently.

I've ended up implementing this one

paolorotolo (approved):
Awesome, looks good!

PR #1110, opened 2023-07-14
Fix several build deprecations

cortinico:
Those build deprecations are appearing on console when building. I'm fixing them up.


paolorotolo (approved):
LGTM, thanks!

PR #1112, opened 2023-07-15
Migrate to ViewPager 2

paolorotolo:
This PR migrates AppIntro from ViewPager to ViewPager2.

## External API Changes:
 * `setCustomTransformer` accepts a `ViewPager2.PageTransformer` instead of `ViewPager.PageTransformer`
 * `setScrollDurationFactor` has been removed since customizing scroll duration is [no longer possible with ViewPager2](https://issuetracker.google.com/issues/122656759). There are some ugly hacks, but if needed we can implement them later.

## Internal development choices:
 * We no longer have a custom view `AppIntroViewPager` internally, because `ViewPager2` is a final class now. Instead, the new class `AppIntroViewPagerController(viewPager2)` handles all the custom logic to set up a ViewPager2 in AppIntro. 

 * ViewPager2 does not allow to override touch events (needed to check drags against intro policy and eventually discard them). However, ViewPager2 supports disabling all user interaction and providing touch events from a fake source (`pager2.fakeDrag()`).

That's why ViewPager2 has been wrapped by a `GestureOverlayView`:
```xml
  <android.gesture.GestureOverlayView
      <androidx.viewpager2.widget.ViewPager2/>
  </android.gesture.GestureOverlayView>
```

To honor Intro custom policy, ViewPager user input has been disabled (`viewPager.isUserInputEnabled = false`) and events from the `GestureOverlay` are instead forwarded to ViewPager2's `fakeDrag` only if policy allows it. If not, they are discarded.

Other changes are just adapting old methods to new ViewPager2 and refactor the code to improve readability, they should be trivial.

PR #1113, opened 2023-07-19
Deprecate setScrollDurationFactor

paolorotolo:
As discussed in #1112, this PR adds a deprecation warning to `setScrollDurationFactor` and removes it from the example intro.

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
@@ -302,10 +302,12 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
      * multiply duration
      * @param factor the new factor that will be applied to the scroll - default: 1
      */
-    @Deprecated("""
+    @Deprecated(
+        """
         Newer versions of ViewPager do not support customizing scroll duration anymore.
         Expect this method to be removed on future versions of AppIntro.
-    """)
+    """
+    )
     protected fun setScrollDurationFactor(factor: Int) {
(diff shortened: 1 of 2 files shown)


appintro/src/main/java/com/github/appintro/AppIntroBase.kt
```diff
@@ -382,6 +388,7 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
      LIFECYCLE
      =================================== */
 
+    @Suppress("DEPRECATION")
```
cortinico, line 391: why do we need this if we removed the `setScrollDurationFactor` call?

cortinico (approved):

appintro/src/main/java/com/github/appintro/AppIntroBase.kt
paolorotolo: I just removed the method from the example, that's a call in the internal `AppIntroBase` activity in the library.

Actually this sets `setScrollDurationFactor(DEFAULT_SCROLL_DURATION_FACTOR)` during intro's `onCreate`. I'm not entirely sure why we need that, since the duration factor is already initialized to `DEFAULT_SCROLL_DURATION_FACTOR`.

But anyway, it was there in the first place and I'm scared to introduce a regression in a bugfix release because probably there's a side effect I'm not considering.

cortinico: Ah sorry I thought it's in the sample code, my bad. It's fine to stay here.

PR #1116, opened 2023-07-24
Fix jitpack build

paolorotolo:
Fix jdk version on jitpack


paolorotolo:
Merging since jetpack build of this commit succeded.

PR #1134, opened 2023-08-26
Persist state in viewmodel for fragments and remove retainInstance. Closes #1050

### LINKED ISSUE #1050: Move away from `Fragment.retainInstance`
## ⚠️ The problem
We're currently using `Fragment.retainInstance` which is deprecated:
https://github.com/AppIntro/AppIntro/blob/3a301e4021e3a81c8684defc63f06bb623afc3b5/appintro/src/main/java/com/github/appintro/AppIntroBaseFragment.kt#L72

We should instead use a ViewModel to store the state of the Fragment and move away from the `onActivityCreated` method as well.

paolorotolo:
This PR:
 - removes fragment deprecated `retainInstance` moving all the state from Fragment in a ViewModel (AppIntroFragmentViewModel)
 - uses ViewModel's saved state to automatically save and restore state when activity is killed by the system
 - to facilitate access to ViewModel's saved state, introduces an helper that sets/gets values from SavedState by delegation
 - since variables are nullable when accessing them in SavedState's key-value store, makes all `Int` nullable instead of relying on the special integer `0`.  This IMHO is also less error-prone since it's easy to distinguish parameters of the Fragment that have been customized by the user (not null) and parameters that have not (null).

Closes #1050


appintro/src/main/java/com/github/appintro/AppIntroFragmentViewModel.kt
```diff
@@ -0,0 +1,56 @@
...
+internal const val ARG_DRAWABLE = "drawable"
+internal const val ARG_BG_COLOR = "bg_color"
+internal const val ARG_BG_COLOR_RES = "bg_color_res"
+internal const val ARG_TITLE_COLOR = "title_color"
+internal const val ARG_TITLE_COLOR_RES = "title_color_res"
+internal const val ARG_DESC_COLOR = "desc_color"
+internal const val ARG_DESC_COLOR_RES = "desc_color_res"
+internal const val ARG_BG_DRAWABLE = "bg_drawable"
+
+class AppIntroFragmentViewModel(state: SavedStateHandle) : ViewModel() {
```
cortinico, line 25: can this also be `internal`?

cortinico (approved):
Amazing cleanup!

paolorotolo:
Done, thanks!

PR #1143, opened 2023-09-19
Fix askForPermission's not required permissions on API>=30. Closes #141

paolorotolo:
After a [platform behavior change](https://developer.android.com/about/versions/11/privacy/permissions#dialog-visibility), `required=false` value ignored on API >= 30 when the user denies a permission multiple times.

This PR checks if the asked permission is not required even when the request is automatically dismissed by the OS as "don't ask again".


paolorotolo:
We should cherry-pick this in `release/6.x` branch and probably publish a bugfix release since this could cause some weird behavior on apps asking for not mandatory permissions on newer Android versions.

cortinico (approved):

leosilberg:
Hi 

Any update when this fix will be released?

PR #1150, opened 2023-10-16
Attach permission-request to specific fragment

### LINKED ISSUE #1149: Ask For Permission attached to specific slide
## ⚠️ Is your feature request related to a problem? Please describe
When dealing with different API-Levels, it might not be nessessary to ask for specific permissions. In those cases, the slide numbers change for permissions on the next slide. This means the developer needs to keep track of the slide position himself.

I would like to see a method that lets the developer to attach a permission to a specifc slide via tags.

```
askForPermissionsOnSlide(arrayOf(Manifest.permission.CAMERA), TAG)
askForPermissionsAfterSlide(arrayOf(Manifest.permission.CAMERA), TAG)
```

If a slide is shown with TAG, then the permission can be requested either on or after the slide is shown.

## 🤚 Do you want to develop this feature yourself?

- [x] Yes
- [ ] No

newhinton:
fixes #1149 

This switches permissionsMap to have a string index.
Then instead of the slide-id we use a tag (or id) to figure out if a request should be made

Todo:
- [ ] Check if slideID works on tagged slides

### DIFF
--- a/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
@@ -155,18 +155,6 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
         }
     }
 
-
-    /**
-     * Adds a new slide at the end of the Intro and associates the permissions with it
-     * @param fragment Instance of Fragment which should be added as slide.
-     * @param permissions - Array of permissions
-     * @param required - Whether the user can change this slide without granting the permissions.
-     */
-    protected fun addSlideWithPermissionRequest(fragment: Fragment, permissions: Array<String>, required: Boolean = true) {
-        addSlide(fragment)
-        askForPermissions(permissions, fragment, required)
-    }
-
     /**
      * Called by the user to associate permissions with slides.
      *
--- a/example/src/main/java/com/github/appintro/example/ui/permissions/PermissionsIntro.kt
+++ b/example/src/main/java/com/github/appintro/example/ui/permissions/PermissionsIntro.kt
@@ -45,8 +45,6 @@ class PermissionsIntro : AppIntro() {
         // Here we ask for Calendar permission on slide 4, but we use the fragment instead of
         // slideNumber
         askForPermissions(arrayOf(Manifest.permission.READ_CALENDAR), calendarPermissionFragment)
-        // you can also use addSlideWithPermissionRequest(fragment: Fragment, permissions: Array<String>, required: Boolean = true)
-        // to achieve the same, without having to remember the fragment yourself
     }
 
 


appintro/src/main/java/com/github/appintro/AppIntroBase.kt
```diff
@@ -121,12 +121,12 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
     private lateinit var backButton: View
     private lateinit var indicatorContainer: ViewGroup
 
-    // Asks the ViewPager for the current slide number. Useful to query the [permissionsMap]
-    @Deprecated("Use currentSlideTag")
+    // Asks the ViewPager for the current slide number. Dont use it to query the [permissionsMap]
```
cortinico, line 124: This field is private though so probably not needed to comment it this way?

appintro/src/main/java/com/github/appintro/AppIntroFragment.kt
```diff
@@ -51,7 +54,8 @@ class AppIntroFragment : AppIntroBaseFragment() {
             @ColorInt descriptionColor: Int? = null,
             @FontRes titleTypefaceFontRes: Int? = null,
             @FontRes descriptionTypefaceFontRes: Int? = null,
-            @DrawableRes backgroundDrawable: Int? = null
+            @DrawableRes backgroundDrawable: Int? = null,
+            id: String? = null
```
cortinico, line 58: I don't think we need the `id` field here at all at this stage

cortinico (changes requested):
I would opt for having only the `Int` and `Fragment` overload and skip the ID approach at least for now.

Also @paolorotolo what's your take on this change?

newhinton:
> I don't think we need the `id` field here at all at this stage

As far as i can tell, fragments don't have an identifier. So this id is required to differentiate between each fragment for the permissionMap.
...though maybe we can abuse java itself. hashCode() _should_ provide some unique string, so that could replace a dedicated ID implementation. Though i feel that this is not as maintainable and easy to understand.

> I would opt for having only the Int and Fragment overload and skip the ID approach at least for now.

Sure, i can make the "String"-one private, but it is still internally required, so it cant be removed.

cortinico:
> As far as i can tell, fragments don't have an identifier. So this id is required to differentiate between each fragment for the permissionMap.

In the method that accepts the `Fragment`, just check that instance against the private `fragment` list and get the `index` from there:
https://github.com/AppIntro/AppIntro/blob/96b48b313f282beb167782f8e9a0ee91430bf5a4/appintro/src/main/java/com/github/appintro/AppIntroBase.kt#L115

There is no need to have extra index at all IMHO

newhinton:
> > As far as i can tell, fragments don't have an identifier. So this id is required to differentiate between each fragment for the permissionMap.
> 
> In the method that accepts the `Fragment`, just check that instance against the private `fragment` list and get the `index` from there:
> 
> https://github.com/AppIntro/AppIntro/blob/96b48b313f282beb167782f8e9a0ee91430bf5a4/appintro/src/main/java/com/github/appintro/AppIntroBase.kt#L115
> 
> There is no need to have extra index at all IMHO

Oh yeah, that is actually better!

newhinton:
I have added a check and a small convenience wrapper that combines the addSlide and askForPermission-calls.

appintro/src/main/java/com/github/appintro/AppIntroBase.kt
```diff
@@ -156,17 +156,36 @@ abstract class AppIntroBase : AppCompatActivity(), AppIntroViewPagerListener {
...
     }
 
+
+    /**
+     * Adds a new slide at the end of the Intro and associates the permissions with it
+     * @param fragment Instance of Fragment which should be added as slide.
+     * @param permissions - Array of permissions
+     * @param required - Whether the user can change this slide without granting the permissions.
+     */
+    protected fun addSlideWithPermissionRequest(fragment: Fragment, permissions: Array<String>, required: Boolean = true) {
```
cortinico, line 166:
```suggestion
    protected fun askForPermissions(permissions: Array<String>, slide: Fragment, required: Boolean = true)
```

cortinico (changes requested):

appintro/src/main/java/com/github/appintro/AppIntroBase.kt
newhinton: I dont quite get what you are trying to do here, this duplicates the askForPermissions-function name and parameters, which doesn't compile?

Do you want me to change the function name or order of parameters?

cortinico: Sorry I haven't seen you added the overload already. This function can go then

appintro/src/main/java/com/github/appintro/AppIntroFragment.kt
```diff
@@ -5,6 +5,7 @@ import androidx.annotation.ColorRes
 import androidx.annotation.DrawableRes
 import androidx.annotation.FontRes
 import com.github.appintro.model.SliderPage
+import java.util.UUID
```
cortinico, line 8: Can you undo this change?

newhinton: Sure, that was accidentially leftover from an older commit.

To the other function:
It is only a tiny wrapper but one that, i imagine, will be useful to most people, since it takes care of the "combining of the permission and slide. Should i really remove it?

appintro/src/main/java/com/github/appintro/AppIntroBase.kt
newhinton: It is only a tiny wrapper but one that, i imagine, will be useful to most people, since it takes care of the "combining of the permission and slide". Should i really remove it?

cortinico: I think we really don't need an extra api where this can be achieved with 2 statements.

newhinton: Done!

PR #1152, opened 2023-10-24
Allow AppIntroFragment to be extended

newhinton:
This "opens" up the AppIntroFragment. If developers can extend the slide-fragments for more functionality. (Eg. changing the layout easily, or adding code specific for each fragment)


cortinico:
We intentionally don't want to open `AppIntroFragment` as you can provide your own Fragment implementations

PR #1179, opened 2024-01-16
Added `setBarMargin`

### LINKED ISSUE #1123: Bottom bar overlapping fragment view
i am using a custom fragment in like this

`
        addSlide(AppListFragment(repository))
`
the issue intro bottom bar is overlapping my fragment  view 

this is version of 

`    implementation 'com.github.AppIntro:AppIntro:6.2.0'
`

358602804_951190229475976_9008040772106030823_n

ArnyminerZ:
As described in #1123 ([comment](https://github.com/AppIntro/AppIntro/issues/1123#issuecomment-1892509134)), I've added the `setBarPadding` function, which adds a padding to the bottom of the slide equal to the bottom bar's height.

Closes #1123


appintro/src/main/java/com/github/appintro/AppIntro.kt
```diff
@@ -199,4 +200,22 @@ abstract class AppIntro : AppIntroBase() {
             bottomSeparator.visibility = View.INVISIBLE
         }
     }
+
+    /**
+     * Adds a margin to the bottom of the content of the slide equal to the bottom bar.
+     * For situations where the bar is not transparent, and you want the content of the slide to
+     * be intractable.
+     *
+     * @param setBarPadding Set : true for margin. false for no margin.
```
cortinico, line 209: Here you mix `margin` and `padding`. Let's stick to one (margin is the correct one here).

appintro/src/main/java/com/github/appintro/AppIntro.kt
```diff
@@ -199,4 +200,22 @@ abstract class AppIntro : AppIntroBase() {
...
+     */
+    fun setBarPadding(setBarPadding: Boolean) {
+        val bottomBar = findViewById<View>(R.id.bottom)
+        val margin = if (setBarPadding)
+            resources.getDimension(R.dimen.appintro_bottombar_height).toInt()
+        else
+            0
+        (bottomBar.layoutParams as? ViewGroup.MarginLayoutParams)
+            ?.setMargins(0, 0, 0, margin)
+            ?.also { bottomBar.requestLayout() }
```
cortinico, line 219: I don't think this works. Specifically you should add marginBottom to `pager_gesture_overlay` and not to `bottom`. Could you add an example to the sample app to make sure this is not broken?

cortinico (changes requested):
Sorry this took longer but I believe we need to adapt the code. Also can you fix the ktlint failure?

appintro/src/main/java/com/github/appintro/AppIntro.kt
ArnyminerZ: True, it has now been fixed, and I've added an example.

example/src/main/java/com/github/appintro/example/ui/custom/MarginsIntro.kt
```diff
@@ -0,0 +1,50 @@
...
+        addSlide(
+            MarginsSlideFragment.newInstance(R.string.margins_no_hide)
+        )
+    }
+
+    override fun onPageSelected(position: Int) {
+        super.onPageSelected(position)
+
+        // Enable the margin only on the last page
+        setBarMargin(position == 2)
```
cortinico, line 38: instead of doing this, could you add a `<CheckBox>` inside the `intro_margins.xml` that toggles the `setBarMargin`?

cortinico (commented):
I've tested it and it works good 👍  can we change the sample a bit and then we'll be able to merge it

README.md
```diff
@@ -379,6 +380,17 @@ setNavBarColor(Color.RED)
 setNavBarColorRes(R.color.red)
 ``` 
 
+### Bottom margin
```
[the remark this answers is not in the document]
cortinico: Yup sorry I got confused

PR #1223, opened 2024-07-14
Adapt build scripts to publish on Maven Central

### LINKED ISSUE #1219: Migrate build scripts to release on Maven Central Repository
We've got `dev.appintro` approved into OSSRH. We should take advantage of our new major `7.0.0` release and configure our build scripts to upload artifacts to Maven Central.

paolorotolo:
Since we've got `dev.appintro` approved into OSSRH, this PR adds the logic in `build.gradle` to publish to `mavenCentral`.

### Testing
Publish a new SNAPSHOT with `gradle publish`.

To test if it's possible to consume it, add Sonatype Snapshot repository:
```kotlin
   maven { url = uri("https://s01.oss.sonatype.org/content/repositories/snapshots/")}
```

Add AppIntro dependency:
```kotlin
   implementation("dev.appintro.appintro:appintro:7.0.0-SNAPSHOT")
```

### Future Developments
Right now, no automation is set. It would be nice to have a GH Action that automatically publishes SNAPSHOTS or releases.

Closes #1219.

### DIFF
--- a/appintro/build.gradle.kts
+++ b/appintro/build.gradle.kts
@@ -3,11 +3,12 @@ plugins {
     alias(libs.plugins.kotlin.android)
     alias(libs.plugins.detekt)
     alias(libs.plugins.ktlint)
-    id("maven-publish")
+    `maven-publish`
+    signing
 }
 
-group = "com.github.AppIntro"
-version = "7.0.0-beta02"
+group = "dev.appintro"
+version = "7.0.0-SNAPSHOT"
 
 android {
     namespace = "com.github.appintro"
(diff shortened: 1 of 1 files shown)


appintro/build.gradle.kts
```diff
@@ -3,11 +3,12 @@ plugins {
     alias(libs.plugins.kotlin.android)
     alias(libs.plugins.detekt)
     alias(libs.plugins.ktlint)
-    id("maven-publish")
+    `maven-publish`
+    signing
```
cortinico, line 7:
```suggestion
    id("maven-publish")
    id("signing")
```

appintro/build.gradle.kts
```diff
@@ -79,38 +80,57 @@ val POM_URL: String by project
 
 publishing {
     publications {
-        register<MavenPublication>("release") {
-            afterEvaluate {
+        afterEvaluate {
```
cortinico, line 83: I think this is not correct, you should keep the original ordering.

paolorotolo:
Sorry for the latest fixes

appintro/build.gradle.kts
```diff
@@ -82,35 +83,54 @@ publishing {
...
                     }
                 }
             }
         }
     }
+
+    repositories {
+        maven {
+            name = "OSSRH"
+            url = uri("https://s01.oss.sonatype.org/content/repositories/snapshots/")
```
cortinico, line 122: Heads up: this will work only for `-SNAPSHOT` releases.
Once we bump to a stable, this repository won't accept the release anymore.

I'd suggest we move to something like this:
https://github.com/gradle-nexus/publish-plugin

cortinico (changes requested):

cortinico:
Sorry for the late review also @paolorotolo

s1204IT:
I would like this feature too.
If it's difficult to generate a build script, there is also the option to upload directly to SonaType.

PR #1337, opened 2026-08-14
Edge to edge support

### LINKED ISSUE #1263: Android 15 Edge to Edge
## 🐛 Describe the bug
Android 15 (sdk 35) force edge to edge display behavior.  See https://developer.android.com/develop/ui/views/layout/edge-to-edge.  When this happens, the Slide's buttons overlap with the bottom menu bar.  And the 

## ✅ Expected behavior
Make the slider bottom buttons not overlap with the devices menu bar.

## 💣 Steps to reproduce
Update target sdk to 35 and run on android 15+ device.

## 📷 Screenshots
image

## 📱 Tech info
 - AppIntro Version: 6.3.1
 - Device: pixel 6a
 - Android OS Version: 15

Slion:
Fix totalSlidesNumber so that it can be used during construction. Make it usable as submodule.
Close AppIntro/AppIntro#1263.

### DIFF
--- a/appintro/build.gradle.kts
+++ b/appintro/build.gradle.kts
@@ -1,6 +1,6 @@
 plugins {
-    alias(libs.plugins.library)
-    alias(libs.plugins.kotlin.android)
+    id("com.android.library")
+    id("org.jetbrains.kotlin.android")
     alias(libs.plugins.detekt)
     alias(libs.plugins.ktlint)
     id("maven-publish")
--- a/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
+++ b/appintro/src/main/java/com/github/appintro/AppIntroBase.kt
@@ -20,9 +20,12 @@ import androidx.appcompat.app.AppCompatDelegate
 import androidx.appcompat.widget.TooltipCompat.setTooltipText
 import androidx.core.app.ActivityCompat
 import androidx.core.content.ContextCompat
+import androidx.core.view.ViewCompat
 import androidx.core.view.WindowCompat
+import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.WindowInsetsCompat.Type.systemBars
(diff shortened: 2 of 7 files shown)


appintro/build.gradle.kts
```diff
@@ -1,6 +1,6 @@
 plugins {
-    alias(libs.plugins.library)
-    alias(libs.plugins.kotlin.android)
+    id("com.android.library")
+    id("org.jetbrains.kotlin.android")
```
cortinico, line 3: please revert this

cortinico (commented):
Would you be able to attach a video of before/after for reference?

Slion:
> Would you be able to attach a video of before/after for reference?

There is a screenshot in the bug. It fixes overlaps with the status bar.

appintro/build.gradle.kts
Slion: It's needed to be able to use it as a submodule otherwise you get errors like:
```
Error resolving plugin [id: 'com.android.library', version: '8.13.0']
> The request for this plugin could not be satisfied because the plugin is already on the classpath with an unknown version, so compatibility cannot be checked.
```

cortinico:
> There is a screenshot in the bug. It fixes overlaps with the status bar.

Ok but can you add before/after screenshots to this PR please?

Slion:
Before:
image

After:
Screenshot_20260825_101904_Fulguris Debug

Slion:
It's been in used for months at [Fulguris ](https://github.com/Slion/Fulguris) so it should be well tested by now. There should be another patch coming up from me about Android TV support.

cortinico:
> Before: image
> 
> After: Screenshot_20260825_101904_Fulguris Debug

Those 2 screenshots are from different app. Please test it against the reference app (or the same app)

